PR #82, opened 2015-01-23
Fix pub global tools.

seaneagan:


devoncarew:
Thanks! lgtm. Can you add yourself to the https://github.com/google/grinder.dart/blob/master/AUTHORS file as well?

devoncarew:
(added)

seaneagan:
Thanks!  I'll add myself to AUTHORS on the next one.

seaneagan:
oh nevermind,  I see you already did.

PR #83, opened 2015-01-29
work on testing for web apps

devoncarew:
Add the ability to drive web tests from a grinder script. This will use either chrome or dartium, depending on whether the files under test are in `test/` or `build/test`.

### DIFF
--- a/bin/test.dart
+++ b/bin/test.dart
@@ -1,0 +1,67 @@
+// Copyright 2015 Google. All rights reserved. Use of this source code is
+// governed by a BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:grinder/grinder.dart';
+
+/**
+ * A small command-line utility to leverage the grinder APIs to make it easy to
+ * run CLI or web app tests.
+ */
+void main(List<String> args) {
+  if (args.length != 1) {
+    print(
+        'usage: pub global activate grinder:test <filepath>');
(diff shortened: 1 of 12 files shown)


bin/test.dart
```diff
@@ -0,0 +1,67 @@
...
+import 'package:grinder/grinder.dart';
+
+/**
+ * A small command-line utility to leverage the grinder APIs to make it easy to
+ * run CLI or web app tests.
+ */
+void main(List<String> args) {
+  if (args.length != 1) {
+    print(
+        'usage: pub global activate grinder:test <filepath>');
```
sunglim: Do you mean `usage: pub global run grinder:test <filepath>`?  (using `run` instead of  `activate`?)

> pub global activate grinder:test
Resolving dependencies... (1.0s)
Could not find package grinder:test at https://pub.dartlang.org.
Depended on by:
- pub global activate

sunglim:
Super! especially printing out webkit insepctor console message from terminal!!

bin/test.dart
devoncarew: I do mean `run` - good catch!

devoncarew:
Thanks! I hope to make it ~as easy to run web tests as command-line ones.

I have some tweaks to do - based on trying to get this working on travis - to fall back to other browsers when chrome isn't available (so, try chrome, chromium, ...).

PR #86, opened 2015-02-05
Update Quiver Dependency

kaendfinger:
Update Dependency on Quiver

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -12,7 +12,7 @@ environment:
 dependencies:
   args: '>=0.10.0 <0.13.0'
   http_server: '>=0.9.0 <0.10.0'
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -12,7 +12,7 @@ environment:
 dependencies:
   args: '>=0.10.0 <0.13.0'
   http_server: '>=0.9.0 <0.10.0'
-  quiver: '>=0.18.0 <0.21.0'
+  quiver: '>=0.18.0 <0.23.0'
```
devoncarew: `quiver` is on `0.21.0+3` right now, so we should limit this to <0.22.0. You are jumping the gun :)

kaendfinger: oh thanks i misread the version. i will fix it tomorrow

PR #87, opened 2015-02-05
use the cli_util's sdk locator method

devoncarew:
Move to using the `cli_util` package's `getSdkDir()` method; @pq.


pq:
w00t!  LGTM. :+1:

PR #89, opened 2015-02-23
Fix #79.

### LINKED ISSUE #79: Annotation to define Tasks
Something like this:

```dart

import 'package:grinder:grinder.dart';

const main = startGrinder;

@GrinderTask()
init() {
  // ...
}

@GrinderTask(deps: const ['init'])
test() {
  // ...
}

```

seaneagan:
@devoncarew 

I didn't deprecate `task` (I can if you want to though), bit I did remove references to it from the README.

### DIFF
--- a/example/declarative.dart
+++ b/example/declarative.dart
@@ -5,10 +5,7 @@ library grinder.example.declarative;
 
 import 'package:grinder/grinder.dart';
 
-void main(List<String> args) {
-  addAnnotatedTasks();
-  startGrinder(args);
-}
+const main = grind;
 
 @Task()
 init(GrinderContext context) {
--- a/example/declarative.dart
+++ b/example/declarative.dart
@@ -7,22 +7,27 @@ import 'package:grinder/grinder.dart';
 
 const main = grind;
 
-@Task()
+@Task(
+    description: 'Initialize stuff.')
 init(GrinderContext context) {
-  context.log("I set things up");
+  context.log("Initializing stuff...");
 }
 
-@Task(depends: const ['init'])
+@Task(
+    depends: const ['init'],
+    description: 'Compile stuff.')
 compile(GrinderContext context) {
-  context.log("I'm compiling now...");
-  //context.fail('woot');
+  context.log("Compiling stuff...");
 }
 
-@Task(depends: const ['compile'])
+@Task(
+    depends: const ['compile'],
+    description: 'Deploy stuff.')
 deploy(GrinderContext context) {
-  context.log("deploying a");
-  context.log("deploying b");
+  context.log("Deploying stuff...");
 }
 
-@Task(name: 'default', depends: const ['deploy'])
+@Task(
+    name: 'default',
+    depends: const ['deploy'])
 _default(GrinderContext context) {}
--- a/example/declarative.dart
+++ b/example/declarative.dart
@@ -1,4 +1,4 @@
-// Copyright 2013 Google. All rights reserved. Use of this source code is
+// Copyright 2015 Google. All rights reserved. Use of this source code is
 // governed by a BSD-style license that can be found in the LICENSE file.
 
 library grinder.example.declarative;
--- a/example/declarative.dart
+++ b/example/declarative.dart
@@ -1,29 +1,31 @@
 // Copyright 2013 Google. All rights reserved. Use of this source code is
 // governed by a BSD-style license that can be found in the LICENSE file.
 
-library grinder_example_1;
+library grinder.example.declarative;
 
 import 'package:grinder/grinder.dart';
 
-void main([List<String> args]) {
-  task('init', init);
-  task('compile', compile, ['init']);
-  task('deploy', deploy, ['compile']);
-  task('default', null, ['deploy']);
-
+void main(List<String> args) {
+  addAnnotatedTasks();
   startGrinder(args);
 }
 
-void init(GrinderContext context) {
+@Task()
+init(GrinderContext context) {
   context.log("I set things up");
 }
 
-void compile(GrinderContext context) {
+@Task(depends: const ['init'])
+compile(GrinderContext context) {
   context.log("I'm compiling now...");
   //context.fail('woot');
 }
 
-void deploy(GrinderContext context) {
+@Task(depends: const ['compile'])
+deploy(GrinderContext context) {
   context.log("deploying a");
   context.log("deploying b");
 }
+
+@Task(name: 'default', depends: const ['deploy'])
+_default(GrinderContext context) {}
--- a/example/declarative.dart
+++ b/example/declarative.dart
@@ -5,7 +5,7 @@ library grinder.example.declarative;
 
 import 'package:grinder/grinder.dart';
 
-const main = grind;
+main(args) => grind(args);
 
 @Task(
     description: 'Initialize stuff.')
--- a/lib/grinder.dart
+++ b/lib/grinder.dart
@@ -1,4 +1,4 @@
-// Copyright 2013 Google. All rights reserved. Use of this source code is
+// Copyright 2015 Google. All rights reserved. Use of this source code is
 // governed by a BSD-style license that can be found in the LICENSE file.
 
 library grinder;
--- a/lib/grinder.dart
+++ b/lib/grinder.dart
@@ -1,41 +1,6 @@
 // Copyright 2013 Google. All rights reserved. Use of this source code is
 // governed by a BSD-style license that can be found in the LICENSE file.
 
-/**
- * A library and tool to drive a command-line build.
- *
- * Grinder build files are entirely specified in Dart code. This allows you to
- * write and debug your build files with the same tools you use for the rest of
- * your project source.
- *
- * Generally, a Grinder implementation will look something like this:
- *     void main([List<String> args]) {
- *       task('init', init);
- *       task('compile', compile, ['init']);
- *       task('deploy', deploy, ['compile']);
- *       task('docs', deploy, ['init']);
- *       task('all', ['deploy', 'docs']);
- *
- *       startGrinder(args);
- *     }
- *
- *     init(GrinderContext context) {
- *       context.log("I set things up");
- *     }
- *
- *     ...
- *
- * Tasks to run are specified on the command line. If a task has dependencies,
- * those dependent tasks are run before the specified task.
- *
- * ## Command-line usage
- *     usage: dart grinder.dart <options> target1 target2 ...
- *
- *     valid options:
- *     --version     print the version of grinder
- *     -h, --help    show targets but don't build
- *     -d, --deps    display the dependencies of targets
- */
 library grinder;
 
 export 'grinder_files.dart';
@@ -44,10 +9,12 @@ export 'grinder_tools.dart';
 import 'dart:async';
 import 'dart:convert' show JSON, UTF8;
 import 'dart:io';
+import 'dart:mirrors';
 
 import 'package:args/args.dart';
 
 import 'src/utils.dart';
+import 'src/get_annotated_tasks.dart';
 
 final Grinder _grinder = new Grinder();
 List<String> _args;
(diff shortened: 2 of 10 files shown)


lib/grinder.dart
```diff
@@ -1,41 +1,6 @@
...
- * those dependent tasks are run before the specified task.
- *
- * ## Command-line usage
- *     usage: dart grinder.dart <options> target1 target2 ...
- *
- *     valid options:
- *     --version     print the version of grinder
- *     -h, --help    show targets but don't build
- *     -d, --deps    display the dependencies of targets
- */
```
seaneagan: Since dartdocs already route the user through the README, I prefer not to duplicate its contents in the main library docs of a package.  Let me know if you disagree.

example/declarative.dart
```diff
@@ -1,29 +1,31 @@
...
 import 'package:grinder/grinder.dart';
 
-void main([List<String> args]) {
-  task('init', init);
-  task('compile', compile, ['init']);
-  task('deploy', deploy, ['compile']);
-  task('default', null, ['deploy']);
-
+void main(List<String> args) {
+  addAnnotatedTasks();
```
seaneagan: It would be cool to have a `package:grinder/main.dart` library:

```void main(List<String> args) {
  addAnnotatedTasks();
  startGrinder(args);
}```

Then one could simply do:

```dart
export `package:grinder/main.dart`;
```

devoncarew: That would certainly be terse :) I know that's a pattern people have used - I've seen it in dart polymer. For me it seems like depending on a side effect - that import (or exporting) a package causes code to be run. And the flow for the user or reader isn't as clear as having an explicit main() method.

The explicit main() is a bit of boilerplate however. So, I'd be ok if you wanted to add a `package:grinder/main.dart` library, I just would want to make it clear that both implicit and explicit ways of starting the library were OK.

devoncarew:
(taking a look at this now)

example/declarative.dart
```diff
@@ -1,29 +1,31 @@
...
   startGrinder(args);
 }
 
-void init(GrinderContext context) {
+@Task()
```
devoncarew: It would be nice to be able to tighten this up, to something like `@task`. Not sure how possible that is given there's already a function named `task`.

example/declarative.dart
```diff
@@ -1,29 +1,31 @@
...
 
-void compile(GrinderContext context) {
+@Task(depends: const ['init'])
+compile(GrinderContext context) {
   context.log("I'm compiling now...");
   //context.fail('woot');
 }
 
-void deploy(GrinderContext context) {
+@Task(depends: const ['compile'])
```
devoncarew: Another annotation scheme:
```
@task
@depends('foo', 'bar');
@description('My nice task description.');
deploy(GrinderContext context) {
...
}
```

It doesn't have the benefit of being all wrapped up in one annotation, but is potentially nicer to read. I don't have a preference but want to bring up other formats.

The new `rpc` package (https://github.com/dart-lang/rpc) uses a single annotation. When writing code that uses it I don't think the `description` is as nice to read. I'd love to be able to just use the dartdoc for the method but that's not available at runtime -

lib/grinder.dart
devoncarew: Can you keep a very terse summary, and point people to the readme? Agree that there's no point to have two places to maintain it.

lib/grinder.dart
```diff
@@ -44,10 +9,12 @@ export 'grinder_tools.dart';
 import 'dart:async';
 import 'dart:convert' show JSON, UTF8;
 import 'dart:io';
+import 'dart:mirrors';
 
 import 'package:args/args.dart';
 
 import 'src/utils.dart';
+import 'src/get_annotated_tasks.dart';
```
devoncarew: nit: we try to keep the imports sorted lexically.

example/declarative.dart
```diff
@@ -1,29 +1,31 @@
...
 
 import 'package:grinder/grinder.dart';
 
-void main([List<String> args]) {
-  task('init', init);
-  task('compile', compile, ['init']);
-  task('deploy', deploy, ['compile']);
-  task('default', null, ['deploy']);
-
+void main(List<String> args) {
```
devoncarew: Re: `addAnnotatedTasks()` - we may want to prefer the annotation mechanism as the best way to run grinder. The programmatic way would exists for people who needed more control over the machinery, but the declarative way would be the 90% case.

So, is there a good way to discover tasks w/o an explicit call to `addAnnotatedTasks()`?
```
main() => startGrinder();

@task
doFoo() {
...
}

@task(depends: 'foo')
doBar() {
...
}
```

lib/src/get_annotated_tasks.dart
```diff
@@ -0,0 +1,55 @@
+
```
devoncarew: Can you add a copyright header?

devoncarew: Also, `add_annotated_tasks.dart` ==> `discover_tasks.dart`?

example/declarative.dart
seaneagan: Agreed.  I wish it could be `@Task`, see http://dartbug.com/13582.

That's the problem I find with `@foo`, it tends to conflict with over names, and it's more difficult to switch between `@foo` and `@Foo(...)` than `@Foo` and `@Foo(...)`.

lib/src/utils.dart
```diff
@@ -37,3 +37,25 @@ class ResettableTimer implements Timer {
...
+      .map((Match match) =>
+          withCapitalization(
+              match.input.substring(match.start, match.end),
+              false))
+      .join('-');
+}
+
+// Upper-case or lower-case the first charater of a String.
+String withCapitalization(String s, bool capitalized) {
+  if(s.isEmpty || capitalized == null) return s;
```
devoncarew: nit: `if(` ==> `if (`

test/annotated_tasks.dart
```diff
@@ -0,0 +1,16 @@
+
```
devoncarew: copyright header

test/get_annotated_tasks_test.dart
```diff
@@ -0,0 +1,69 @@
+// Copyright 2013 Google. All rights reserved. Use of this source code is
```
devoncarew: nit: copyright year

devoncarew:
Sorry for the delay - I made a pass through. There are some open-ended comments in-line -

example/declarative.dart
seaneagan: That's what I was going for with `export 'package:grinder/main.dart';`.  I also considered what you did above i.e. calling `addAnnotatedTasks()` within `startGrinder()`, but if we do want to allow people to opt out of annotation scanning, then we will might still want to allow them to use `startGrinder`.  So another idea is to add this method to grinder:

```dart
void grind(List<String> args) {
  addAnnotatedTasks();
  startGrinder();
}
```

Then the 90% use case can be:

```dart
main(args) => grind(args);
```
or even:

```dart
const main = grind;
```

I definitely agree having a single entry-point for most use cases is nice, then we can change what it does without it being a breaking change for most.

Just let me know which option you like best, and I can implement it.

example/declarative.dart
seaneagan: Thats not bad, but I personally do prefer having a single annotation, maybe I should update the examples to have a similar formatting though?

```dart
@Task(
  depends: const ['foo', 'bar'],
  description: 'My nice task description.')
deploy(GrinderContext context) {
...
}
```
Once we get http://dartbug.com/4046, then we can get rid of `const ` at least.

re: using doc-comments, absolutely, I want to do the same in unscripted:

https://github.com/seaneagan/unscripted/issues/11

but it's blocked on:

http://dartbug.com/15704

seaneagan: Also, I think `depends` -> `deps` and `description` -> `help` would be nice, but then we should change the `GrinderTask` constructor args and properties as well.

example/declarative.dart
seaneagan: Ok, my plan is to rename `addAnnotatedTasks` to `_discoverTasks` (private for now, we can always expose it later), and add the `grind` method mentioned above.  Then if people need to programmatically add tasks they can just do that before calling `grind`:

```dart
main(args) {
  addTask(someTask);
  grind(args);
}

@Task(...)
...
```

test/get_annotated_tasks_test.dart
devoncarew: And generally we don't update the copyright year when modifying a file - we just need to make sure that new files have the copyright (and that we update the year for that file).

example/declarative.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright 2015 Google. All rights reserved. Use of this source code is
+// governed by a BSD-style license that can be found in the LICENSE file.
+
+library grinder.example.declarative;
+
+import 'package:grinder/grinder.dart';
+
+const main = grind;
```
devoncarew: This code is correct, but I think the example will read simpler if it's 
```
main() => grind();
```

And an individual user can do the funky assignment in their scripts if they choose.

example/declarative.dart
```diff
@@ -0,0 +1,28 @@
...
+  //context.fail('woot');
+}
+
+@Task(depends: const ['compile'])
+deploy(GrinderContext context) {
+  context.log("deploying a");
+  context.log("deploying b");
+}
+
+@Task(name: 'default', depends: const ['deploy'])
```
devoncarew: We haven't published the `default` change yet. What do you think about having a `default` bool property on a task, instead of depending on a naming convention? It would mean we wouldn't have to have a no-op method to hang an annotation off.

devoncarew:
> Ok, my plan is to rename addAnnotatedTasks to _discoverTasks (private for now, we can always expose it later), and add the grind method mentioned above. Then if people need to programmatically add tasks they can just do that before calling grind:

I think we're running up against the limits of the size of PR that can be comfortably reviewed in github. `grind` might take a named param to optionally disable annotation scanning (it would scan by default).

> Also, I think depends -> deps and description -> help would be nice, but then we should change the GrinderTask constructor args and properties as well.

I'm ok with depends and description but don't think it's critical. We'd need a lot less `const`s and braces if depends took a string, and the task names were space or comma separated. We'd have to be confident that users could discover that they list multiple tasks using commas in the string rather than strings in a dart list.

devoncarew:
Let me know when you're done making changes; we can land an iterate in new PRs.

devoncarew:
An example of the above: `depends: "foo, bar"` rather than `depends: const ["foo", "bar"]`.

lib/src/get_annotated_tasks.dart
seaneagan: Renamed `addAnnotatedTasks` to `_discoverTasks`, left `getAnnotatedTasks` et al. as is.

example/declarative.dart
seaneagan: Plan implemented.

lib/grinder.dart
seaneagan: Done.

example/declarative.dart
seaneagan: Updated the formatting of existing annotation examples to be more readable.

example/declarative.dart
seaneagan: Went the equally terse `const main = grind` instead.

example/declarative.dart
seaneagan: Yeah, that'll make it easier to tweak the example code to add programmatic tasks too.

Done.

lib/grinder.dart
seaneagan: Done.  I linked to the pub package page, since I didn't know how to link the README page in the dartdocs, and the package page is the most likely README to match the version of the docs they are looking at.

seaneagan:
Done with changes.

I like `grind(scan: true)` or similar, in which case I think we could deprecate `startGrinder`.  It would be a backwards compatible change.

I considered allowing the dependencies to link back to the dependent task declaration:

```dart
@Task()
init(...) {}

@Task(depends: const [init])
compile(...) {}
```

Then rename refactoring works etc... and as a slight bonus you avoid the quotation marks.  It would still need to allow String deps as well for depending on programmatically added tasks.

example/declarative.dart
seaneagan: Since there can only be one default task, maybe it should be a `Grinder.defaultTask` property, and then we can have an `@DefaultTask()` annotation, and then either add an `isDefault: false` parameter to `addTask`, or a top-level `defaultTask` setter for programmatic use cases.  If you try to set the default task twice, it should fail fast.

You'll still need a separate task declaration when you want your default task to run multiple other tasks.

```dart
@DefaultTask(depends: const [deploy, docs])
all() {}
```

seaneagan:
@devoncarew I filed #91, #92, and #93 for outstanding issues that can be addressed in subsequent PRs once we get this merged.

devoncarew:
Merging now - thanks for bearing with me on all the back-and-forth!

seaneagan:
Thanks!  And thanks for all the feedback!

PR #97, opened 2015-02-25
Fix #92

### LINKED ISSUE #92: Allow default task to have any name
Instead of:

```dart
@Task(name: 'default', depends: const ['deploy', 'docs'])
_default() {}
```

```dart
@DefaultTask(depends: const ['deploy', 'docs'])
all() {}
```

Do we want to remove the `name` parameter from `@Task` then?  So far I only needed it because `default` is a keyword in dart.

For programmatic use cases, we'll probably also want a `Grinder.defaultTask` property which can only be assigned to once.  Maybe also expose this as a top-level `defaultTask` property.  Or add a `bool asDefault: false` param to `addTask`.

seaneagan:
@devoncarew

### DIFF
--- a/test/all.dart
+++ b/test/all.dart
@@ -4,14 +4,14 @@
 library all_test;
 
 import 'cli_test.dart' as cli_test;
-import 'discover_tasks_test.dart' as get_annotated_tasks_test;
+import 'discover_tasks_test.dart' as discover_tasks_test;
 import 'grinder_test.dart' as grinder_test;
 import 'grinder_files_test.dart' as grinder_files_test;
 import 'grinder_tools_test.dart' as grinder_tools_test;
 
 main() {
   cli_test.main();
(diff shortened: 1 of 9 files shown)


test/all.dart
```diff
@@ -4,7 +4,7 @@
 library all_test;
 
 import 'cli_test.dart' as cli_test;
-import 'get_annotated_tasks_test.dart' as get_annotated_tasks_test;
+import 'discover_tasks_test.dart' as get_annotated_tasks_test;
```
devoncarew: get_annotated_tasks_test ==> discover_tasks_test

devoncarew:
lgtm! You probably want to update the library name of `lib/src/discover_tasks.dart` to match the file name.

seaneagan:
Updated the library names.

Issue #78, opened 2015-01-22
Use Zone value to access GrinderContext so it doesn't have to be passed around everywhere.


seaneagan:
* More convenient
* Less grinder lock in effect

seaneagan:
I do something similar with `now` in https://pub.dartlang.org/packages/clock.

devoncarew:
Makes sense. It's something to evaluate w/ a bunch of other breaking API changes.

seaneagan:
@devoncarew What do you think about allowing task callbacks to either take a GrinderContext arg or not for awhile, but deprecate taking it, and eventually remove?

devoncarew:
Yeah, sounds good to me. I like the idea of using Zones to have global-ish variables available to each running task. And your technique is backwards compatible, so it could be done at any time.

I do want to try and get a good picture of all potential API breaking changes, so we can make sure they all work well together, and they can be batched up into as few major version releases as possible.

sethladd:
+1 to the idea of using Zone for a "thread local" context object.

devoncarew:
https://github.com/seaneagan/clock/blob/master/lib/src/zoned_value.dart

devoncarew:
done -

PR #99, opened 2015-02-25
add an 'init' command

### LINKED ISSUE #76: command to initialize tool/grinder.dart
It would be cool to do:

```shell
stagehand ...
den install grinder
pub run grinder:init
```

To initialize a grinder project.  `init` would probably not need to be in `executables`, it would initialize a default `tool/grind(er).dart` script.

devoncarew:
Fix #76; @seaneagan

### DIFF
--- a/bin/init.dart
+++ b/bin/init.dart
@@ -1,21 +1,34 @@
 // Copyright 2015 Google. All rights reserved. Use of this source code is
 // governed by a BSD-style license that can be found in the LICENSE file.
 
+library grinder.init;
+
 import 'dart:io';
 
 /**
  * A command to create a simple, starting `tool/grind.dart` build script.
  */
 void main(List args) {
+  if (!new File('pubspec.yaml').existsSync()) {
+    _fail('This script must be run from the project root.');
+  }
+
   File file = new File('tool${Platform.pathSeparator}grind.dart');
 
   if (file.existsSync()) {
-    print('Error: ${file.path} already exists.');
-    exit(1);
+    _fail('Error: ${file.path} already exists.');
   }
+
+  // Create `tool` if it does not already exist.
+  new Directory('tool').createSync();
 
   file.writeAsStringSync(_grindSampleSource);
   print('Wrote ${file.path}!');
+}
+
+void _fail(String message) {
+  print(message);
+  exit(1);
 }
 
 final String _grindSampleSource = '''
@@ -26,7 +39,7 @@ main(args) => grind(args);
 @Task()
 init(GrinderContext context) => defaultInit(context);
 
-@Task(depends: const ['init'])
+@DefaultTask(depends: const ['init'])
 build(GrinderContext context) {
   Pub.build(context);
 }
--- a/bin/init.dart
+++ b/bin/init.dart
@@ -1,0 +1,36 @@
+// Copyright 2015 Google. All rights reserved. Use of this source code is
+// governed by a BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+/**
+ * A command to create a simple, starting `tool/grind.dart` build script.
+ */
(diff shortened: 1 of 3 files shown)


bin/init.dart
```diff
@@ -0,0 +1,36 @@
...
+ */
+void main(List args) {
+  File file = new File('tool${Platform.pathSeparator}grind.dart');
+
+  if (file.existsSync()) {
+    print('Error: ${file.path} already exists.');
+    exit(1);
+  }
+
+  file.writeAsStringSync(_grindSampleSource);
```
seaneagan: I think `writeAsStringSync` will create parent directories (`tool` here) if they don't exist, but the docs don't say.

bin/init.dart
```diff
@@ -0,0 +1,36 @@
+// Copyright 2015 Google. All rights reserved. Use of this source code is
+// governed by a BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+/**
+ * A command to create a simple, starting `tool/grind.dart` build script.
+ */
+void main(List args) {
+  File file = new File('tool${Platform.pathSeparator}grind.dart');
```
seaneagan: I think this should either validate that we are at the root of a pub package (`pubspec.yaml` exists) or try to climb the file system until a `pubspec.yaml` is found and produce a path relative to that.

bin/init.dart
```diff
@@ -0,0 +1,36 @@
...
+  if (file.existsSync()) {
+    print('Error: ${file.path} already exists.');
+    exit(1);
+  }
+
+  file.writeAsStringSync(_grindSampleSource);
+  print('Wrote ${file.path}!');
+}
+
+final String _grindSampleSource = '''
```
seaneagan: Probably should add a library name e.g. `library $packageName.tool.grind;`.

seaneagan: If so, then we should update the examples too.

bin/init.dart
devoncarew: I think we should validate and fail otherwise; I'll add that.

bin/init.dart
```diff
@@ -0,0 +1,36 @@
...
+import 'package:grinder/grinder.dart';
+
+main(args) => grind(args);
+
+@Task()
+init(GrinderContext context) => defaultInit(context);
```
seaneagan: There is support for using properties:

```dart
@Task()
const init = defaultInit;
```

What do you think?

bin/init.dart
devoncarew: Ah, I'll explicitly create the `tool` dir first.

readme.md
```diff
@@ -15,8 +15,11 @@ source.
 
 ## Getting Started
 
-Your grinder build file should reside at `tool/grind.dart`, and may look
-something like:
+Your grinder build file should reside at `tool/grind.dart`. You can use grinder
+to create a simple, starting build script. Run `pub run grinder:init`; this will
```
seaneagan: nit: might be worth putting `pub run grinder:init` on its own line, just so it stands out.

bin/init.dart
seaneagan: scratch that, it doesn't help the uninitiated user figure out how to define arbitrary tasks.

seaneagan:
lgtm!

readme.md
devoncarew: done

devoncarew:
- updated readme
- added a library name to the `init.dart` script
- check for running from the project root

bin/init.dart
seaneagan: Once #78 is fixed, it may make more sense though.

devoncarew: I like simple and explicit for new users, at the expense of being slightly more verbose.

PR #105, opened 2015-02-27
Fix #91.  Fix #104.

### LINKED ISSUE #91: Support rename refactoring for task dependencies.
If the dependency links directly back to the declaration:

```dart
@Task()
init() {}

@Task(depends: const [init])
compile() {}
```

Then `init` can be renamed safely and conveniently using editor rename refactoring.

Also, `init` is shorter than `'init'`.

Will still need to support e.g. `'init'` as well for programmatic use cases.

### LINKED ISSUE #104: Ability to export tasks from other libraries
```dart
// useful_tasks.dart

@Task(description: 'Useful.')
useful(GrinderContext context) {}

@Task(description: 'Also useful.', depends: const [useful])
alsoUseful(GrinderContext context) {}
```

```dart
// tool/grind.dart';
import 'package:useful_tasks/useful_tasks.dart';
export 'package:useful_tasks/useful_tasks.dart';

main(args) => grind(args);

@Task(depends: const [useful, alsoUseful])
build(GrinderContext context) {}
```

You cannot depend on tasks you do not export.

If you don't actually want to call the tasks by themselves, then you just call them within your task instead of depending on them:

```dart
// tool/grind.dart';
import 'package:useful_tasks/useful_tasks.dart';

main(args) => grind(args);

@Task()
build() {
  useful(context);
  alsoUseful(context);
  // ...
}
```

seaneagan:
@devoncarew 

Sorry for conflating these two, I didn't realize until almost done with 91 that 104 was separable.

I didn't include any examples of how to use 104 yet.  I think that can be a separate PR.  Thinking maybe can refactor some of the common built in tasks to be able to be exported, and then doing that from the examples or in the build file produced by `pub run grinder:init`.


example/declarative.dart
```diff
@@ -14,14 +14,14 @@ init(GrinderContext context) {
 }
 
 @Task(
-    depends: const ['init'],
+    depends: const [init],
```
devoncarew: In another PR, we may want to add to our declarative example (or in the readme) and example of a camelCase method name and the string name to refer to it by on the CLI.

devoncarew:
lgtm!

example/declarative.dart
seaneagan: Yep, I think we could show the resulting output of `grind -h` for the example pubspec, which will show how the resulting task names and descriptions.

Issue #103, opened 2015-02-26
Use doc comments to define task description


seaneagan:
```dart
/// Compiles stuff.
@Task()
compile() {
  // ...
}
```

instead of:

```dart
@Task(
  description: 'Compiles stuff.')
compile() {
  // ...
}
```

This would allow producing nice dartdocs for reusable task libraries!

Of course, it's not currently possible, see: http://dartbug.com/15704.

devoncarew:
+1!

devoncarew:
@seaneagan, @bwilkerson

I'm finding the new task format a little hard to read:
```dart
@Task(
    depends: const ['init'],
    description: 'Compile stuff.')
compile(GrinderContext context) {
  context.log("Compiling stuff...");
}
```

But, I very much like the move towards annotations in general. I think a more-easily-read format would be something like:
```dart
/// Compile stuff.
@Task(depends: init)
compile(GrinderContext context) {
  context.log("Compiling stuff...");
}

or

/// Compile stuff.
@task
@depends(init)
compile(GrinderContext context) {
  context.log("Compiling stuff...");
}
```

But those re-worked formats require us having the ability to pull dartdoc out of our code at runtime. Brian, what's the feasibility of using the analyzer to do this? Given a path to a library file and a symbol name (`compile`), could we extract the dartdoc for that symbol reliably and performantly?

bwilkerson:
> Given a path to a library file and a symbol name (compile), could we extract the dartdoc for that symbol reliably and performantly?

If the name is defined in the library file (rather than a part), then the only thing you need to do is read, scan, and parse the file, then find the symbol in the file. If it might be in a part, then you might have to repeat those steps after resolving the part URI's. That should be reasonably performant.

seaneagan:
That'd be awesome if we can use the analyzer.

Regarding splitting into multiple annotations e.g. `@task` and `@depends`, I agree it looks pretty nice.  However, I'm concerned about losing auto-complete when typing `@Task(`.  But if we did go with it, and using the analyzer doesn't work for some reason, this isn't too bad:

```dart
@task('Compile stuff.')
@depends(init)
compile(GrinderContext context) {
  context.log("Compiling stuff...");
}
```

but then it would have to be `@task()` instead of `@task` until dart allows that.

devoncarew:
Ah, so use `@task(...)` as both the task marker _and_ the annotation for documentation. I like how it reads.

devoncarew:
[...]

PR #107, opened 2015-03-02
Fix #106.

### LINKED ISSUE #106: Separate varargs annotation for task dependencies
```dart
@Task('A nice task')
@Depends(dep1, dep2, 'dep3')
niceTask() {
  // ...
}
```

as discussed in #103.

seaneagan:
@devoncarew 

I'd still like to try #103 in a later PR, but if it doesn't work out, I'm happy with this as the final state of annotations.

### DIFF
--- a/example/declarative.dart
+++ b/example/declarative.dart
@@ -7,22 +7,19 @@ import 'package:grinder/grinder.dart';
 
 main(args) => grind(args);
 
-@Task(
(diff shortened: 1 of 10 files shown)


devoncarew:
lgtm! The task declarations read very well now. Even long descriptions would look good; they could use triple quote strings, and look nearly as good as dartdoc.

PR #111, opened 2015-03-02
Fix #94.

### LINKED ISSUE #94: Deprecate `task`
Now that we have declarative task definitions (#79), I don't think we need the `task` function.  I think `addTask` will be more appropriate for programmatic use cases.  I think the biggest programmatic use case will be sharing task definitions across packages/libraries, so something like `addTask(someImportedTask)`.

seaneagan:
@devoncarew


lib/grinder.dart
```diff
@@ -55,6 +55,7 @@ set defaultTask(GrinderTask v) {
  * Any dependencies of the task, that need to run before it, should be passed
  * in via [depends].
  */
+@deprecated
```
devoncarew: What about `@Deprecated('Prefer instead using the task annotations')`?

devoncarew:
We should also update the library name for the renamed `example/declarative.dart` file.

lgtm!

lib/grinder.dart
seaneagan: done.

seaneagan:
I just removed the library name altogether to avoid people copying it into their projects and having an incorrect library name.  That matches what we do with the init command.

PR #115, opened 2015-03-11
Fix #114

### LINKED ISSUE #114: Deprecate startGrinder
We now have `grind` which does first discovers annotated tasks, and then does what startGrinder does.  If we want to allow opting out of annotated task scanning we can do that via `grind` in #93.

seaneagan:
@devoncarew

### DIFF
--- a/lib/src/cli.dart
+++ b/lib/src/cli.dart
@@ -1,3 +1,5 @@
+// Copyright 2015 Google. All rights reserved. Use of this source code is
+// governed by a BSD-style license that can be found in the LICENSE file.
 
 library grinder.src.cli;
 
--- a/lib/src/cli.dart
+++ b/lib/src/cli.dart
@@ -1,0 +1,119 @@
+
+library grinder.src.cli;
+
+import 'dart:async';
+import 'dart:convert' show JSON, UTF8;
+import 'dart:io';
+
+import 'package:args/args.dart';
+
+import 'singleton.dart';
+import 'utils.dart';
+import '../grinder.dart';
+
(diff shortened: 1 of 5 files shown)


lib/src/cli.dart
```diff
@@ -0,0 +1,119 @@
...
+import 'dart:io';
+
+import 'package:args/args.dart';
+
+import 'singleton.dart';
+import 'utils.dart';
+import '../grinder.dart';
+
+// This version must be updated in tandem with the pubspec version.
+const String APP_VERSION = '0.6.6+3';
```
devoncarew: Any reason to move this constant out of `lib/grinder.dart`? (if we're moving this, we should update the comment in the pubspec file).

lib/src/cli.dart
```diff
@@ -0,0 +1,119 @@
+
```
devoncarew: Can you add a copyright header?

lib/src/singleton.dart
```diff
@@ -0,0 +1,6 @@
+
```
devoncarew: copyright header

lib/src/cli.dart
devoncarew: Hmm, nevermind - I see this was a library private variable in the main lib, and is now a public var in a private library. So, a bit of cleanup :)

Let's adjust https://github.com/google/grinder.dart/blob/master/pubspec.yaml#L2 to match.

devoncarew:
lgtm, w/ 2 copyright headers and one small comment update

Issue #108, opened 2015-03-02
Analyze grinder file produced by `pub run grinder:init`


seaneagan:
As long as the output of `pub run grinder:init` is static, it'd be nice to keep it in an actual dart file, so that we can get analysis warnings/errors in it without having to actually run.  I don't think it's possible to get the source of an imported library using `dart:mirrors`, so I think we'd have to wait for http://dartbug.com/21020 or use some workaround for that.

However, if we foresee the output becoming dynamic (e.g. adding a library name), then we'd need to add a test to actually run the command and run the analyzer on the output.

We could run other tests on the produced grinder file as well, not sure what.

devoncarew:
+1 to analyzing the `init` output as part of the build. We do something like this for stagehand, and it has caught issues. Nice to have safety nets :)

PR #117, opened 2015-03-13
verify the generated grind script

devoncarew:
Fix https://github.com/google/grinder.dart/issues/108

@seaneagan

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -42,14 +42,17 @@ Future testsBuildWeb(GrinderContext context) {
 @Task('Analyze the generated grind script')
 @Depends(init)
 analyzeInit(GrinderContext context) {
+  final sep = Platform.pathSeparator;
+
   Directory dir = new Directory('init_temp');
 
   try {
     dir.createSync();
-    File pubspec = new File('init_temp/pubspec.yaml');
+    File pubspec = new File('init_temp${sep}pubspec.yaml');
     pubspec.writeAsStringSync('name: foo', flush: true);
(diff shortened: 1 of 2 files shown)


tool/grind.dart
```diff
@@ -2,37 +2,55 @@
...
 }
+
+@Task('Analyze the generated grind script')
+@Depends(init)
+analyzeInit(GrinderContext context) {
+  Directory dir = new Directory('init_temp');
+
+  try {
+    dir.createSync();
+    File pubspec = new File('init_temp/pubspec.yaml');
```
seaneagan: Might be good to use the `path` package.  I'm usually stuck on a windows machine, and I *think* it will work anyways with these paths, but not 100% sure.

tool/travis.sh
```diff
@@ -16,6 +16,9 @@ pub global run tuneup check --ignore-infos
 # Run the tests.
 dart test/all.dart
 
+# Verify that the generated grind script analyzes well.
+dart tool/grind.dart analyze-init
```
seaneagan: Eventually would be nice to do `pub run grinder:grind analyze-init` or even just `grind`, see #118.

seaneagan:
lgtm!

tool/grind.dart
devoncarew: Makes sense; I moved over to using Platform.pathSeparator for now.

tool/travis.sh
devoncarew: +1. I'm not particularly concerned w/ any problems associated with self-hosting; if it breaks in a PR we should always just fix the issue w/ that PR.

Issue #36, opened 2014-09-23
Consider delete() instead of deleteEntity()


sethladd:
`deleteEntity(new Directory('path'))` is a bit verbose. What do you think about `delete(string_or_directory)` ?

seaneagan:
I like overloads because because once we get [union types](http://dartbug.com/4938) the type checking can be reinstated.

seaneagan:
I think this is done.  Filed #138 to track removing deprecated things later on.

Issue #102, opened 2015-02-26
Use /// doc-comments instead of /** */


seaneagan:
Can either do this manually, or wait until the [linter supports it](https://github.com/dart-lang/linter/issues/38).

devoncarew:
I'm old school in that I prefer `/** ... */`, but if much of the code is using `///` then consistency is more important -

devoncarew:
Closing - I think the resolution is to convert over as files are edited -

PR #120, opened 2015-03-13
simplify the file manipulation code a bit

devoncarew:
partially address #36 

- `copyFile` and `copyDirectory` and deprecated in favor of a new `copy` method
- `deleteEntity` is deprecated in favor of a new `delete` method
- also added a new `Path` class, in order to experiment with an easier way to essentially add `name`, `copy`, and `delete` methods to file system entities

### DIFF
--- a/lib/grinder_files.dart
+++ b/lib/grinder_files.dart
@@ -14,6 +14,8 @@ import 'package:glob/glob.dart';
 
 import 'grinder.dart';
 
+final String _sep = Platform.pathSeparator;
+
 // TODO: add files to a set
 
 // TODO: union sets?
@@ -97,13 +99,113 @@ class FileSet {
   }
 }
 
+abstract class Path {
+  factory Path(FileSystemEntity entity) => new _EntityPath(entity);
+  factory Path.str(String p) => new _StringPath(p);
+
+  Path._();
+
+  String get name;
+
+  String get path;
+
+  bool get isDirectory;
+
+  bool get exists;
+
+  FileSystemEntity get entity;
+
+  void copyPath(Directory destDir) => copy(entity, destDir);
+
+  void deletePath() => delete(entity);
+
+  Path join([arg0, String arg1, String arg2, String arg3, String arg4,
+    String arg5, String arg6, String arg7, String arg8, String arg9]) {
+    List args = [];
+
+    if (arg0 is List) {
+      args = arg0;
+    } else if (arg0 is String) {
+      _addNonNull(args, arg0);
+      _addNonNull(args, arg1);
+      _addNonNull(args, arg2);
+      _addNonNull(args, arg3);
+      _addNonNull(args, arg4);
+      _addNonNull(args, arg5);
+      _addNonNull(args, arg6);
+      _addNonNull(args, arg7);
+      _addNonNull(args, arg8);
+      _addNonNull(args, arg9);
+    }
+
+    if (args.isEmpty) {
+      return this;
+    } else {
+      return new Path.str('${path}${_sep}${args.join(_sep)}');
+    }
+  }
+
+  File returnAsFile() => new File(path);
+
+  Directory returnAsDirectory() => new Directory(path);
+
+  String toString() => path;
+}
+
+class _EntityPath extends Path {
+  final FileSystemEntity _entity;
+
+  _EntityPath(this._entity) : super._();
+
+  FileSystemEntity get entity => _entity;
+
+  bool get exists => _entity.existsSync();
+
+  bool get isDirectory => _entity is Directory;
+
+  String get name => fileName(_entity);
+
+  String get path => _entity.path;
+}
+
+class _StringPath extends Path {
+  final String _path;
+
+  _StringPath(this._path) : super._();
+
+  FileSystemEntity get entity {
+    final FileSystemEntityType type = FileSystemEntity.typeSync(_path);
+
+    if (type == FileSystemEntityType.FILE) {
+      return new File(_path);
+    } else if (type == FileSystemEntityType.DIRECTORY) {
+      return new Directory(_path);
+    } else if (type == FileSystemEntityType.LINK) {
+      return new Link(_path);
+    } else {
+      return null;
+    }
+  }
+
+  bool get exists =>
+      FileSystemEntity.typeSync(_path) != FileSystemEntityType.NOT_FOUND;
+
+  bool get isDirectory => FileSystemEntity.isDirectorySync(_path);
+
+  String get name {
+    int index = _path.lastIndexOf(_sep);
+    return index != -1 ? _path.substring(index + 1) : null;
+  }
+
+  String get path => _path;
+}
+
 /**
  * Return the last segment of the file path.
  */
 String fileName(FileSystemEntity entity) {
   String name = entity.path;
-  int index = name.lastIndexOf(Platform.pathSeparator);
-
+  int index = name.lastIndexOf(_sep);
   return (index != -1 ? name.substring(index + 1) : name);
 }
 
@@ -122,19 +224,18 @@ String fileExt(FileSystemEntity entity) {
  */
 String baseName(FileSystemEntity entity) {
   String name = entity.path;
-  int index = name.lastIndexOf(Platform.pathSeparator);
-
+  int index = name.lastIndexOf(_sep);
   return (index != -1 ? name.substring(0, index) : null);
 }
 
 File joinFile(Directory dir, List<String> files) {
-  String pathFragment = files.join(Platform.pathSeparator);
-  return new File("${dir.path}${Platform.pathSeparator}${pathFragment}");
+  String pathFragment = files.join(_sep);
+  return new File("${dir.path}${_sep}${pathFragment}");
 }
 
 Directory joinDir(Directory dir, List<String> files) {
-  String pathFragment = files.join(Platform.pathSeparator);
-  return new Directory("${dir.path}${Platform.pathSeparator}${pathFragment}");
+  String pathFragment = files.join(_sep);
+  return new Directory("${dir.path}${_sep}${pathFragment}");
 }
 
 /**
@@ -142,10 +243,10 @@ Directory joinDir(Directory dir, List<String> files) {
  * given path to a platform dependent path.
  */
 File getFile(String path) {
-  if (Platform.pathSeparator == '/') {
+  if (_sep == '/') {
     return new File(path);
   } else {
-    return new File(path.replaceAll('/', Platform.pathSeparator));
+    return new File(path.replaceAll('/', _sep));
   }
 }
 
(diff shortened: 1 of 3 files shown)


devoncarew:
@sethladd @seaneagan 

- `copyXXX` ==> `copy`
- `deleteEntity ` ==> `delete`
- and, added a new class to streamline file manipulation. It uses param type overloads to make the paths easy to create, and once created, the objects make it easy to manipulate file system resources.

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+    if (arg is FileSystemEntity) return arg.path;
+    if (arg is Path) return arg.path;
+    throw new ArgumentError('expected a FileSystemEntity or a String');
+  }
+}
+
+/**
  * Return the last segment of the file path.
  */
 String fileName(FileSystemEntity entity) {
```
seaneagan: What's the advantage of this over `basename` from the `path` package?

lib/grinder_files.dart
```diff
@@ -122,19 +322,18 @@ String fileExt(FileSystemEntity entity) {
  */
 String baseName(FileSystemEntity entity) {
```
seaneagan: same question for `dirname` from the `path` package.

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+//   */
+//  Path get absolute {
+//    // TODO:
+//  }
+
+  bool get isDirectory => FileSystemEntity.isDirectorySync(_path);
+  bool get isFile => FileSystemEntity.isFileSync(_path);
+  bool get isLink => FileSystemEntity.isLinkSync(_path);
+
+  /**
```
seaneagan: Per #102 can we make this /// ?  Or we can just wait until the linter supports batch fixes.

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+   * Synchronously create the file. See also [File.create].
+   *
+   * If [recursive] is false, the default, the file is created
+   * only if all directories in the path exist. If [recursive] is true, all
+   * non-existing path components are created.
+   */
+  void createFile({bool recursive: false}) {
+    asFile.createSync(recursive: recursive);
```
seaneagan: return the result of `asFile`?

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+
+  static Path get cwd => new Path(Directory.current);
+
+  final String _path;
+
+  /**
+   * Create a new [Path]. The [entity_or_path] parameter can be a
+   * [FileSystemEntity] or a [String].
+   */
+  Path(entity_or_path) : _path = _coerce(entity_or_path);
```
seaneagan: entity_or_path => entityOrPath

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+  /**
+   * Creates a temporary directory in the system temp directory. See
+   * [Directory.systemTemp] and [Directory.createTempSync]. If [prefix] is
+   * missing or null, the empty string is used for [prefix].
+   */
+  static Path createSystemTemp([String prefix]) {
+    return new Path(Directory.systemTemp.createTempSync(prefix));
+  }
+
+  static Path get cwd => new Path(Directory.current);
```
seaneagan: `cwd` => `current` for consistency?

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
 }
 
 /**
+ * A class to make it easier to manipulate file system entites. Once paths or
+ * entites are converted into `Path`s, they can be easily copied, deleted,
+ * joined, and their name retrieved.
```
seaneagan: Can we just depend on the path manipulation functionality in the `path` package for this?

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
 }
 
 /**
+ * A class to make it easier to manipulate file system entites. Once paths or
+ * entites are converted into `Path`s, they can be easily copied, deleted,
```
seaneagan: Much of the non-path-manipulation functionality here is available on FileSystemEntity, but dart:io has no implementation of FileSystemEntity which can abstractly represent any type (file, directory, link).  Might be cool to have `Path` extend `FileSystemEntity`.  Just an idea.

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+   * You would call this instead of [entity] when the file system entity does
+   * not yet exist.
+   */
+  Link get asLink => new Link(path);
+
+  /**
+   * Copy the the entity to the given destination. Return the newly created
+   * [Path].
+   */
+  Path copy(Path destDir) {
```
seaneagan: Maybe also (or instead) allow `String` or `Directory`?

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+    return new Path(destDir).join(name);
+  }
+
+  /**
+   * Delete the entity at the path.
+   */
+  void delete() => _deleteImpl(entity);
+
+  /**
+   * Synchronously create the file. See also [File.create].
```
seaneagan: File.create => File.createSync

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+   * non-existing path components are created.
+   */
+  void createLink(String target, {bool recursive: false}) {
+    asLink.createSync(target, recursive: recursive);
+  }
+
+  /**
+   * Join the given path elements to this path, and return a new [Path] object.
+   */
+  Path join([arg0, String arg1, String arg2, String arg3, String arg4,
```
seaneagan: Should `arg0` be required?

lib/grinder_files.dart
seaneagan: Also, `fileName` seems like the wrong name, since it also works for directories.

lib/grinder_files.dart
```diff
@@ -122,19 +322,18 @@ String fileExt(FileSystemEntity entity) {
  */
 String baseName(FileSystemEntity entity) {
   String name = entity.path;
-  int index = name.lastIndexOf(Platform.pathSeparator);
-
+  int index = name.lastIndexOf(_sep);
   return (index != -1 ? name.substring(0, index) : null);
 }
 
 File joinFile(Directory dir, List<String> files) {
```
seaneagan: If we add `Path`, maybe this is easy enough: `new Path(directory).join(...).asDirectory`?

lib/grinder_files.dart
```diff
@@ -154,43 +353,56 @@ File getFile(String path) {
...
-  File destFile = joinFile(destDir, [fileName(srcFile)]);
+void copy(FileSystemEntity entity, Directory destDir, [GrinderContext context]) {
+  return _copyImpl(entity, destDir, context);
+}
 
-  if (!destFile.existsSync() ||
-      srcFile.lastModifiedSync() != destFile.lastModifiedSync()) {
+void _copyImpl(FileSystemEntity entity, Directory destDir, [GrinderContext context]) {
+  if (entity is Directory) {
     if (context != null) {
```
seaneagan: I think this should go in `copy` since the usage from `Path` doesn't care about `GrinderContext`s.

test/grinder_files_test.dart
```diff
@@ -159,4 +159,81 @@ main() {
...
+  group('grinder.files Path', () {
+    Path temp;
+
+    setUp(() {
+      temp = Path.createSystemTemp();
+    });
+
+    tearDown(() => temp.delete());
+
+    test('Path.entity', () {
```
seaneagan: Path.entity => entity since Path is already included in the parent group name?

seaneagan: Oh, I see it's for testing passing an entity to the constructor.  Maybe 'Path.entity' => 'Path(entity)'

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+
+    if (args.isEmpty) {
+      return this;
+    } else {
+      return new Path('${path}${_sep}${args.join(_sep)}');
+    }
+  }
+
+  bool operator ==(other) {
+    return other is Path ? path == other.path : false;
```
seaneagan: I'd do: `other is Path && path == other.path`.

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+      _addNonNull(args, arg6);
+      _addNonNull(args, arg7);
+      _addNonNull(args, arg8);
+      _addNonNull(args, arg9);
+    }
+
+    if (args.isEmpty) {
+      return this;
+    } else {
+      return new Path('${path}${_sep}${args.join(_sep)}');
```
seaneagan: If you do `var paths = [path];` above, instead of `var args = [];`, then this can change to just `new Path(paths.join(_sep))`.

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+    } else if (type == FileSystemEntityType.DIRECTORY) {
+      return new Directory(_path);
+    } else if (type == FileSystemEntityType.LINK) {
+      return new Link(_path);
+    } else {
+      return null;
+    }
+  }
+
+  bool get exists {
```
seaneagan: nit: Add doc comment since behavior is different from FileSystemEntity.exists.

lib/grinder_files.dart
seaneagan: Or keep the functionality, but implement it with the `path` package to decrease likelihood of bugs.

lib/grinder_files.dart
```diff
@@ -98,12 +100,210 @@ class FileSet {
...
+  }
+
+  /**
+   * Synchronously create the link. See also [Link.create].
+   *
+   * If [recursive] is false, the default, the file is created
+   * only if all directories in the path exist. If [recursive] is true, all
+   * non-existing path components are created.
+   */
+  void createLink(String target, {bool recursive: false}) {
```
seaneagan: `copy` takes a `Path`, while this takes a `String`.  Can we settle on one or allow multiple?  Also, what about `FileSystemEntity`?

seaneagan:
@devoncarew sorry for the delay!  Done reviewing.

lib/grinder_files.dart
devoncarew: sgtm, done

lib/grinder_files.dart
devoncarew: makes sense

devoncarew:
Thanks for the review! I'm going to work my way through the comments. I find it really hard to process bigger CL's with the github PR interface...

lib/grinder_files.dart
devoncarew: Hmm. Yeah, what I really want to do is just add additional methods to `FileSystemEntity`. Having `Path` implement that would work. The name `Path` generally also means something else in a dart context (as in, a library prefix for the `path` library). So a different class name like `Entity` might be disambiguating.

lib/grinder_files.dart
devoncarew: Yup, makes sense, to build on its testing.

test/grinder_files_test.dart
devoncarew: done

lib/grinder_files.dart
devoncarew: done

lib/grinder_files.dart
devoncarew: sure, will change these over -

lib/grinder_files.dart
devoncarew: Right now I'm thinking that you'll convert things into Paths, operate on them as Paths, and then convert them out, back into file system entities.

lib/grinder_files.dart
devoncarew: done

lib/grinder_files.dart
devoncarew: sgtm, done

lib/grinder_files.dart
devoncarew: I converted this param over to a `Path` for consistency.

lib/grinder_files.dart
devoncarew: I'm not sure one way or another; `join()` would be the same as `join([])`.

lib/grinder_files.dart
devoncarew: cool, done

lib/grinder_files.dart
devoncarew: done

lib/grinder_files.dart
devoncarew: This method has been around for a while, and hasn't changed for this PR. I'd have been _much_ happier if FileSystemEntity has just had a `name` property...

lib/grinder_files.dart
devoncarew: This could definitely be re-implemented in terms of Path.

lib/grinder_files.dart
devoncarew: makes sense; done

PR #133, opened 2015-03-23
use zones to store the context

devoncarew:
@seaneagan, fix issue https://github.com/google/grinder.dart/issues/78 - use zones to make variables available to running tasks, rather than passing params around.


lib/src/singleton.dart
```diff
@@ -4,5 +4,20 @@
...
+
+class _NoopContext implements GrinderContext {
+  Grinder get grinder => null;
+
+  GrinderTask get task => null;
+
+  // TODO: implement fail
+  void fail(String message) { }
+
+  // TODO: implement log
```
devoncarew: Fast fail here? Or just print to stdout?

lib/src/utils.dart
```diff
@@ -99,3 +99,33 @@ Map<Symbol, DeclarationMirror> resolveExportedDeclarations(LibraryMirror library
 getFirstMatchingAnnotation(DeclarationMirror decl, bool test(annotation)) =>
   decl.metadata.map((mirror) => mirror.reflectee).firstWhere(test,
       orElse: () => null);
+
+/// A simple way to expose a default value that can be overridden within zones.
+class ZonedValue<T> {
```
devoncarew: From `clock`...

readme.md
```diff
@@ -34,20 +34,20 @@ import 'package:grinder/grinder.dart';
...
 
 @Task('Compile stuff.')
 @Depends(init)
-compile(GrinderContext context) {
-  context.log("Compiling stuff...");
+compile() {
+  log("Compiling stuff...");
 }
 
 @DefaultTask('Deploy stuff.')
```
devoncarew: What do you think about `@Task('foo', isDefault: true)`?

seaneagan: The reason I went with `@DefaultTask` is to make it stick out as much as possible from the other tasks, since:

* It's the most frequently run (thus important) task.
* There can only be one, so it should be easy to scan to verify that.
* To avoid confusion that the task has the same name as the method, as with non-default tasks.

For example:

```dart
@Task('''
This is an awesome task.
You should run it all of the time.''', isDefault: true)
awesome() {
  // ...
}
```

would be easy to misread as just a regular `awesome` task.  That seems less likely with:

```dart
@DefaultTask('''
This is an awesome task.
You should run it all of the time.''')
awesome() {
  // ...
}
```

devoncarew: OK, makes sense. It struck me as odd - when scanning that script - that not all the tasks started with `@Task...`. But I see your reasoning.

lib/grinder_files.dart
```diff
@@ -386,16 +386,13 @@ void _copyImpl(FileSystemEntity entity, Directory destDir, [GrinderContext conte
 /**
  * Delete the given file entity reference.
  */
-void delete(FileSystemEntity entity, [GrinderContext context]) {
-  _deleteImpl(entity, context);
+void delete(FileSystemEntity entity, [GrinderContext _context]) {
```
seaneagan: Since `delete` hasn't been released yet, we can remove the `_context` parameter right?

lib/grinder_tools.dart
```diff
@@ -139,10 +139,10 @@ Future runProcessAsync(GrinderContext context, String executable,
  * A default implementation of an `init` task. This task verifies that the grind
  * script is executed from the project root.
  */
-void defaultInit(GrinderContext context) {
+void defaultInit([GrinderContext context]) {
```
seaneagan: For the APIs that take a GrinderContext parameter which is now optional, can we mark it as deprecated?

lib/grinder_files.dart
devoncarew: done

lib/grinder_tools.dart
devoncarew: We can doc the param as deprecated. There are unfortunately many API methods which take `GrinderContext` as their first parameter. This means we can't change it to an optional param. We can either:

- keep the param, but no longer refer to it in the method, or
- remove the param completely. It would be a breaking change from the `0.6.0` version -

lib/grinder.dart
```diff
@@ -137,19 +147,26 @@ class GrinderTask {
...
       {this.taskFunction, this.depends : const [], this.description});
 
   /**
    * This method is invoked when the task is started. If a task was created with
-   * a [TaskFunction], that function will be invoked by this method.
+   * a [Function], that function will be invoked by this method.
    */
-  dynamic execute(GrinderContext context) {
-    return (taskFunction != null) ? taskFunction(context) : null;
+  dynamic execute(GrinderContext _context) {
```
seaneagan: I think the _context param should be optional and deprecated now.

lib/grinder_files.dart
```diff
@@ -386,16 +386,13 @@ void _copyImpl(FileSystemEntity entity, Directory destDir, [GrinderContext conte
...
 }
 
-void _deleteImpl(FileSystemEntity entity, [GrinderContext context]) {
+void _deleteImpl(FileSystemEntity entity) {
   if (entity.existsSync()) {
-    if (context != null) {
-      context.log('deleting ${entity.path}');
-    }
-
+    log('deleting ${entity.path}');
```
seaneagan: So now the usage from Path will log as well.  I like that.  Do the same in _copyImpl?

lib/src/discover_tasks.dart
```diff
@@ -66,15 +66,18 @@ class TaskDiscovery {
...
+      Function taskFunction;
 
       if (decl is VariableMirror ||
           (decl is MethodMirror && decl.isGetter)) {
         taskFunction = owner.getField(decl.simpleName).reflectee;
       } else if (decl is MethodMirror &&
                  decl.isRegularMethod) {
-        taskFunction = (GrinderContext context) =>
-            owner.invoke(decl.simpleName, [context]);
+        if (decl.parameters.isNotEmpty) {
```
seaneagan: Check for at least one positional param to avoid passing a grinder context to something which only has named parameters?

lib/src/singleton.dart
```diff
@@ -4,5 +4,20 @@
 library grinder.src.singleton;
 
 import '../grinder.dart';
+import 'utils.dart';
 
 final Grinder grinder = new Grinder();
+
+final ZonedValue zonedContext = new ZonedValue(new _NoopContext());
+
+class _NoopContext implements GrinderContext {
```
seaneagan: If we create new GrinderContexts for each task / zone then it will be possible to use the GrinderContext outside of the associated task execution window.  Do we want to allow that?  If no, what if we have GrinderContext be a singleton (still accessed via `context`) and then we just store the task in each zone:

```dart
final GrinderContext context = new GrinderContext();

class GrinderContext {
  GrinderTask get task => zonedTask.value;
  Grinder get grinder => grinder; // from singleton.dart
  void log(String msg) {
    if (task == null) throw new GrinderException('GrinderContext.log can only be called during execution of a task');
    // log ...
  }

  // similar for fail ...
}
```

lib/src/singleton.dart
seaneagan: Failing fast could help catch errors, and we can always stop throwing later without it being a breaking change.

Same for fail.

tool/grind.dart
```diff
@@ -21,19 +21,19 @@ void analyze(GrinderContext context) {
...
-Future testsWeb(GrinderContext context) {
+Future testsWeb() {
   return Tests.runWebTests(context, directory: 'web', htmlFile: 'web.html');
 }
 
 @Task()
 @Depends(init)
-Future testsBuildWeb(GrinderContext context) {
+Future testsBuildWeb() {
   return Pub.buildAsync(context, directories: ['web']).then((_) {
```
seaneagan: I guess we can't make `GrinderContext` params from methods which already have named arguments optional since dart doesn't allow a method to have both named and positional optionals.  I guess we can't provide a non-breaking upgrade path for that.  I think maybe those methods are relatively new though, and thus wouldn't have that many users yet, so should we ahead and remove the params now?

lib/grinder.dart
```diff
@@ -73,6 +72,17 @@ Future startGrinder(List<String> args) {
   return handleArgs(args);
 }
 
+// Zone variables.
+
+/// Get the [GrinderContext] for the currently executing task.
+GrinderContext get context => zonedContext.value;
+
+/// Log an informational message to Grinder's output.
+void log(String message) => context.log(message);
```
seaneagan: If we're exposing `log` and `fail` at the top-levl, we could expose `task` (possibly renamed to `currentTask` or `GrinderTask.current`) and `grinder` (or `Grinder.current`), and then deprecate GrinderTask altogether.  What do you think?

seaneagan:
@devoncarew done reviewing.  I have some open questions inline, but I like where this is going!

lib/grinder_tools.dart
seaneagan: Since I think most (all?) of the changes so far for 0.7.0 are non-breaking, I vote we keep the required params for now (and ignore their values), and since it's impossible to deprecate the required params, we just doc-comment that they will be removed soon (and that they are ignored), and then in the next breaking version after 0.7.0 we remove the all the context params (even the optional ones).

lib/grinder_files.dart
seaneagan: Unless you are wanting to eventually factor out `Path` into some other package, so it can be used outside grinder scripts as well.

lib/grinder_tools.dart
seaneagan: I take that back, why make users go through two separate migrations.  I think we should move the parameters that can't be made optional now.  Then they can update all their code when moving to 0.7.0, but at the same time for the newly optional params, they are not required to update yet if they don't want to, but will get a nice deprecation message if we mark as such.  I'd also be fine with just removing the newly optional ones too.

seaneagan:
@devoncarew I'm happy to address my remaining comments in a follow up PR if you want, up to you.

lib/src/singleton.dart
devoncarew: +1 to fast fail

lib/grinder_tools.dart
devoncarew: I'm considering just removing all `context` params, and making `0.7.0` a big breaking change. I don't think it'll be a burden on the users; there aren't any semantic changes in the breaking changes, just removed args.

lib/src/discover_tasks.dart
devoncarew: sounds good; will do

lib/src/singleton.dart
devoncarew: Sounds good; I like not having the dangling contexts.

devoncarew:
Sorry for the delay in responding! I'll spend a few hours cleaning this up; we can then land and iterate. I definitely appreciate the comments; it is making the API and impl better :)

lib/grinder_tools.dart
devoncarew: > I think we should move the parameters that can't be made optional now. Then they can update all their code when moving to 0.7.0, but at the same time for the newly optional params, they are not required to update yet if they don't want to

+1 :)

devoncarew:
Merging this. I'd like to use other PRs to look at:
- making sure a context fast fails if it's used outside of it's task
- setting up temporary contexts for the tests (instead if `test(`, something like `zonedTest(`)

PR #139, opened 2015-03-27
Update args dependency

kaendfinger:
cc @devoncarew 

After this, a release should probably be published, because of latest release of Dart.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -10,7 +10,7 @@ environment:
   sdk: '>=1.0.0 <2.0.0'
 
 dependencies:
-  args: '>=0.12.0 <0.14.0'
+  args: '>=0.10.0 <0.14.0'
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -10,7 +10,7 @@ environment:
   sdk: '>=1.0.0 <2.0.0'
 
 dependencies:
-  args: '>=0.10.0 <0.13.0'
+  args: '>=0.12.0 <0.14.0'
```
devoncarew: I'm assuming we do require at least `0.12.0`? I'd like to keep the constraints as broad as possible.

kaendfinger: Ok, I'll expand it, because we don't require 0.12.

kaendfinger:
@devoncarew Fixed

devoncarew:
Thanks!

PR #144, opened 2015-03-28
deprecate defaultInit()

### LINKED ISSUE #137: Consider baking in defaultInit to every grinder script
Does every grinder script want to do what defaultInit is currently doing which is checking that the script is run from the project root?  If so, it would be nice to run that check automatically before any tasks are run, and obviated the need for everyone creating an `init` task.

devoncarew:
fix #137 

Move the functionality of `defaultInit()` into the grinder startup sequence (`handleArgs()`). Deprecate defaultInit. I kept a sample `init` task in the bootstrap grind.dart script we generate in order to demonstrate how to specify task dependencies.

@seaneagan

### DIFF
--- a/lib/grinder.dart
+++ b/lib/grinder.dart
@@ -57,9 +57,9 @@ void task(String name, [Function taskFunction, List<String> depends = const []])
 /// recognized options such as --help.
 ///
 /// If a task fails, throw a [GrinderException] and runs no further tasks.
-Future grind(List<String> args) => new Future(() {
+Future grind(List<String> args, {bool verifyLocation: true}) => new Future(() {
   discoverTasks(grinder, currentMirrorSystem().isolate.rootLibrary);
-  return handleArgs(args);
+  return handleArgs(args, verifyLocation: verifyLocation);
 });
 
 /**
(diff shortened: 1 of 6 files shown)


lib/grinder.dart
```diff
@@ -57,9 +57,9 @@ void task(String name, [Function taskFunction, List<String> depends = const []])
 /// recognized options such as --help.
 ///
 /// If a task fails, throw a [GrinderException] and runs no further tasks.
-Future grind(List<String> args) => new Future(() {
+Future grind(List<String> args, {bool verifyLocation: true}) => new Future(() {
```
seaneagan: I think including "root" would help self-document:

verifyRoot
verifyRootInvoke
verifyRootCwd

lib/grinder.dart
```diff
@@ -57,9 +57,9 @@ void task(String name, [Function taskFunction, List<String> depends = const []])
 /// recognized options such as --help.
 ///
 /// If a task fails, throw a [GrinderException] and runs no further tasks.
```
seaneagan: Explain `verify...` in doc comment.

seaneagan:
lgtm w/ a name suggestion!

lib/grinder.dart
devoncarew: sgtm; perhaps `verifyProjectRoot`

PR #160, opened 2015-04-09
Expose ProcessResult fields (stdout, stderr, exitCode)

seaneagan:
@devoncarew

### DIFF
--- a/lib/grinder.dart
+++ b/lib/grinder.dart
@@ -59,7 +59,7 @@ void task(String name, [Function taskFunction, List<String> depends = const []])
 /// If [verifyProjectRoot] is true, grinder will verify that the script is being
 /// run from a project root.
 ///
-/// If a task fails, throw a [GrinderException] and runs no further tasks.
+/// If a task fails, throws an exception and runs no further tasks.
 Future grind(List<String> args, {bool verifyProjectRoot: true}) => new Future(() {
   discoverTasks(grinder, currentMirrorSystem().isolate.rootLibrary);
   return handleArgs(args, verifyProjectRoot: verifyProjectRoot);
@@ -68,7 +68,7 @@ Future grind(List<String> args, {bool verifyProjectRoot: true}) => new Future(()
 /**
  * Start the build process. This should be called at the end of the `main()`
  * method. If there is a task failure, this method will halt task execution and
- * throw a [GrinderException].
+ * throw an exception.
  */
 @Deprecated('Use `grind` instead.')
 Future startGrinder(List<String> args, {bool verifyProjectRoot: true}) {
--- a/lib/grinder.dart
+++ b/lib/grinder.dart
@@ -59,7 +59,7 @@ void task(String name, [Function taskFunction, List<String> depends = const []])
 /// If [verifyProjectRoot] is true, grinder will verify that the script is being
 /// run from a project root.
 ///
-/// If a task fails, throws an exception and runs no further tasks.
+/// If a task fails, it throws and runs no further tasks.
 Future grind(List<String> args, {bool verifyProjectRoot: true}) => new Future(() {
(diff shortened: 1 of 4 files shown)


seaneagan:
Rather than return a full ProcessResult on success (and failure?), this exposes stdout on success, and exitCode/stderr on failure (via an exception).  I didn't think it was important enough to expose the `pid` of a process that is already complete.

lib/grinder_tools.dart
```diff
@@ -84,53 +85,57 @@ void runProcess(String executable,
...
+
     // Handle stdout.
-    process.stdout.listen((List<int> data) {
-      if (!quiet) {
-        String str = new String.fromCharCodes(data).trimRight();
-        if (str.isNotEmpty) log(str);
-      }
-    });
+    var broadcastStdout = process.stdout.asBroadcastStream();
+    var stdoutLines = _toLineStream(broadcastStdout);
```
seaneagan: This way the line breaks in the grinder scripts output matces that of the run process.

lib/grinder_tools.dart
```diff
@@ -155,13 +160,15 @@ class Pub {
    * even if the pubspec.lock file is up-to-date with respect to the
    * pubspec.yaml file.
    */
-  static void get({bool force: false, String workingDirectory}) {
+  static String get({bool force: false, String workingDirectory}) {
```
seaneagan: Not sure how useful it is to access the stdout of some of these commands.  If there are any that you would prefer to leave `void` just let me know which ones.  Would it be at all useful to instead return a `bool` here for whether `pub get` was actually run?

lib/grinder.dart
```diff
@@ -59,7 +59,7 @@ void task(String name, [Function taskFunction, List<String> depends = const []])
 /// If [verifyProjectRoot] is true, grinder will verify that the script is being
 /// run from a project root.
 ///
-/// If a task fails, throw a [GrinderException] and runs no further tasks.
+/// If a task fails, throws an exception and runs no further tasks.
```
devoncarew: `it throws`

lib/grinder_tools.dart
```diff
@@ -84,53 +85,57 @@ void runProcess(String executable,
...
 
-  if (result.stderr != null && !result.stderr.isEmpty) {
+  if (result.stderr != null && result.stderr.isNotEmpty) {
     log(result.stderr);
   }
 
   if (result.exitCode != 0) {
-    throw new GrinderException(
-        "${executable} failed with a return code of ${result.exitCode}");
+    throw new ProcessException(executable, result.exitCode, result.stderr);
```
devoncarew: Should we also return the stdout with this exception? Some CLIs may not put all the info relevant to triaging a failure into stderr.

lib/grinder_tools.dart
devoncarew: I don't have a strong preference here. `String` is consistent. For several of the operations the caller really won't care about the result. For those I would prefer `String` or `void` over `bool`.

lib/grinder_tools.dart
seaneagan: sgtm.  At that point we're only missing the `pid` from `ProcessResult`, so should we just attach one of those to the exception, or keep the individual fields?  I'd say keep the individual fields.

Also let me know if you don't like the name `ProcessException`.  I initially had `NonZeroExitCodeException` which is the most accurate, but a bit long.  Could also be `ProcessRunException` or given #161 maybe `RunException`.

lib/grinder.dart
seaneagan: done.

lib/grinder_tools.dart
seaneagan: Changed back the ones I didn't think were useful to `void`.

PR #162, opened 2015-04-09
Fix #161

### LINKED ISSUE #161: Shorten runProcess[Async]/runDartScript
In the spirit of `log`, `fail`, `copy`, `delete`, might be nice to shorten up `runProcess`, `runProcessAsync`, and `runDartScript`.  `run` / `runAsync` seems clear enough to me without `Process`:

`run('dart', ['--version'])`

For `runDartScript` we could crib off of e.g. `Pub.run`, `Dart2js.compile`, etc. and have a `Dart.run`.

`Dart.run('tool/foo.dart')`

 @devoncarew

seaneagan:
@devoncarew

### DIFF
--- a/lib/grinder_tools.dart
+++ b/lib/grinder_tools.dart
@@ -8,7 +8,6 @@
 library grinder.tools;
 
 import 'dart:async';
-import 'dart:convert';
 import 'dart:io';
 import 'dart:math' as math;
 
@@ -16,9 +15,13 @@ import 'package:cli_util/cli_util.dart' as cli_util;
 import 'package:which/which.dart';
 
 import 'grinder.dart';
+import 'src/run.dart' as r;
+import 'src/run_utils.dart';
 import 'src/utils.dart';
 import 'src/_mserve.dart';
 import 'src/_wip.dart';
+
+export 'src/run.dart';
 
 final Directory BIN_DIR = new Directory('bin');
 final Directory BUILD_DIR = new Directory('build');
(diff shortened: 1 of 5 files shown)


lib/src/run.dart
```diff
@@ -0,0 +1,143 @@
+
```
devoncarew: Can you add a copyright header?

lib/src/run_utils.dart
```diff
@@ -0,0 +1,18 @@
+
```
devoncarew: copyright header

lib/grinder_tools.dart
```diff
@@ -16,10 +15,14 @@ import 'package:cli_util/cli_util.dart' as cli_util;
 import 'package:which/which.dart';
 
 import 'grinder.dart';
+import 'src/run.dart' as r;
```
devoncarew: Hmm, is there a bit better import prefix for this? `r` seems brief...

lib/src/run_utils.dart
seaneagan: done

lib/src/run.dart
seaneagan: done

lib/grinder_tools.dart
seaneagan: Changed to `run_lib`.  I originally had `run`, but it conflicted with `Pub.run` etc.

devoncarew:
lgtm!

Issue #153, opened 2015-03-30
Refactor PubApplication


seaneagan:
@devoncarew This is for the `PubApplication` class added in #71.  This doesn't need to block 0.7.0, but labeled as such so we can at least discuss it first:

* I liked the original shorter name `PubApp` better, no need to spell out `Application`.
* `isInstalled()` should be a getter, possibly renamed to `isActivated`.
* It should allow running any cli app from the package, not just the same named one.  The name `PubApplication` suggests it should be a constructor parameter (e.g. `executable: 'foo'`), and not a `run` parameter, in which case `isInstalled` should check the `pub global list` output to see if the individual script is installed ([once that is supported](http://dartbug.com/20641)).  But it's then weird to `activate()` or `update()` an individual script, so maybe change their name to `activatePackage` / `updatePackage`.  Probably also need to add `packageName` along-side `appName`.
* `pub run` has some advantages over `pub global run`.  Particularly matching the version of the package containing the script with the version used by the local package.  It should be supported as well.  Perhaps it can be an interface with (private) implementations for both global and local packages.  It can have an `isGlobal` property which is set by `PubApp.global` and `PubApp.local` constructors.  But `PubApp.activate()` doesn't make sense for local apps ([at least yet](https://docs.google.com/document/d/1lvbzyivCejhTuHLG4WI3YVThbQqJhPUrwajwTC-sXWU)), so maybe make it private for now, and just rely on it being done as part of `PubApp.run`?  Or just make it a no-op.  `PubApp.update` could potentially run `pub upgrade` (in which case, rename to `upgrade`?).  `PubApp.isInstalled` should check whether the script's containing package is in the pubspec.lock, and use the version found there to check the pub cache for the individual script existing in /bin.

devoncarew:
> I liked the original shorter name PubApp better, no need to spell out Application.

+1. I had some trouble differentiating the name from the tuple returned from PubGlobal.list(), which I've now just called `AppVersion`.

> isInstalled() should be a getter, possibly renamed to isActivated.

+1

> It should allow running any cli app from the package, not just the same named one.

What if the `run()` method takes an optional named param, for the script name? So the simple case of:
```
PubApp grinder = new PubApp('grinder');
grinder.run('foo');
```
works, and the less common case of:
```
PubApp grinder = new PubApp('grinder');
grinder.run('foo', script: 'init'); // or 'grinder:ini'?
```
also works.

> pub run has some advantages over pub global run.

[...]

PR #164, opened 2015-04-10
refactor PubApplication to PubApp

devoncarew:
Refactoring of `PubApplication`, in-line with comments in https://github.com/google/grinder.dart/issues/153.

@seaneagan

### DIFF
--- a/lib/grinder_tools.dart
+++ b/lib/grinder_tools.dart
@@ -276,11 +276,16 @@ class Pub {
 
 /// Access the `pub global` commands.
 class PubGlobal {
+  Set<String> _activatedPackages;
+
   PubGlobal._();
 
   /// Install a new Dart application.
-  void activate(String package) {
-    runProcess(_sdkBin('pub'), arguments: ['global', 'activate', package]);
+  void activate(String packageName, {bool force: false}) {
+    if (force || !isActivated(packageName)) {
+      runProcess(_sdkBin('pub'), arguments: ['global', 'activate', packageName]);
+      _activatedPackages.add(packageName);
+    }
   }
 
   /// Run the given installed Dart application.
@@ -313,7 +318,15 @@ class PubGlobal {
 
   /// Returns whether the given Dart application is installed.
   bool isActivated(String packageName) {
-    return list().any((AppVersion app) => app.name == packageName);
+    if (_activatedPackages == null) _initActivated();
+    return _activatedPackages.contains(packageName);
+  }
+
+  void _initActivated() {
+    if (_activatedPackages == null) {
+      _activatedPackages = new Set();
+      _activatedPackages.addAll(list().map((appVer) => appVer.name));
+    }
   }
 }
 
@@ -335,11 +348,10 @@ abstract class PubApp {
 
   bool get isActivated;
 
-  /// Install the application (run `pub global activate`).
-  void activate();
-
-  /// Install the application or update it to the lastest version.
-  void update();
+  /// Install the application (run `pub global activate`). Setting [force] to
+  /// try will force the activation of the package even if it is already
+  /// installed.
+  void activate({bool force: false});
 
   /// Run the application. If the application is not installed this command will
   /// first activate it.
@@ -353,32 +365,18 @@ abstract class PubApp {
 }
 
 class _PubGlobalApp extends PubApp {
-  bool _activated = false;
-
   _PubGlobalApp(String packageName) : super._(packageName);
 
   bool get isGlobal => true;
 
-  bool get isActivated {
-    if (_activated) return true;
-    _activated = Pub.global.isActivated(packageName);
-    return _activated;
-  }
+  bool get isActivated => Pub.global.isActivated(packageName);
 
-  void activate() {
-    if (!_activated) {
-      Pub.global.activate(packageName);
-      _activated = true;
-    }
-  }
-
-  void update() {
-    Pub.global.activate(packageName);
-    _activated = true;
-  }
+  void activate({bool force: false}) =>
+      Pub.global.activate(packageName, force: force);
 
   String run(List<String> arguments, {String script, String workingDirectory}) {
-    if (!_activated && !isActivated) activate();
+    activate();
+
     return Pub.global.run(packageName,
         script: script,
         arguments: arguments,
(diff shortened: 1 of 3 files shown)


lib/grinder_tools.dart
```diff
@@ -310,50 +310,98 @@ class PubGlobal {
...
-  bool isInstalled(String packageName) {
+  bool isActivated(String packageName) {
     return list().any((AppVersion app) => app.name == packageName);
   }
 }
 
 /// A Dart command-line application, installed via `pub global activate`.
-class PubApplication {
+abstract class PubApp {
   final String appName;
```
seaneagan: `appName` -> `package` (or `packageName`)?

lib/grinder_tools.dart
```diff
@@ -310,50 +310,98 @@ class PubGlobal {
...
 
-  bool _installed = false;
+  PubApp._(this.appName);
 
   /// Create a new reference to a pub application; [appName] is the same as the
   /// package name.
-  PubApplication(this.appName);
+  factory PubApp.global(String appName) => new _PubGlobalApp(appName);
```
seaneagan: Filed #165 for adding support for git and path packages.

lib/grinder_tools.dart
```diff
@@ -310,50 +310,98 @@ class PubGlobal {
...
 
-  bool isInstalled() {
-    if (_installed) return true;
-    _installed = Pub.global.isInstalled(appName);
-    return _installed;
-  }
+  /// Create a new reference to a pub application; [appName] is the same as the
+  /// package name.
+  factory PubApp.local(String appName) => new _PubLocalApp(appName);
```
seaneagan: IMO best practice for many apps will be local due to matching the version of the package used locally, so I wonder if it should just be e.g. `new PubApp('test')..run()`?  My only issue with local apps is [this](https://docs.google.com/document/d/1lvbzyivCejhTuHLG4WI3YVThbQqJhPUrwajwTC-sXWU/edit).  But maybe driving more folks to use them will help draw attention to that problem.

lib/grinder_tools.dart
```diff
@@ -310,50 +310,98 @@ class PubGlobal {
...
-  /// Install the application or update it to the lastest version.
   void update() {
     Pub.global.activate(appName);
-    _installed = true;
+    _activated = true;
   }
 
-  String toString() => appName;
+  String run(List<String> arguments, {String script, String workingDirectory}) {
+    if (!_activated && !isActivated) activate();
```
seaneagan: Similar to `Pub.get(force: true)` I think we should move this logic to `Pub.global.activate` and allow e.g. `Pub.global.activate('stagehand`, force: true)` and `new PubApp.global('stagehand')..activate(force: true)..run(...)`.  Then we can also remove `PubApp.update`.

lib/grinder_tools.dart
```diff
@@ -310,50 +310,98 @@ class PubGlobal {
...
+}
+
+class _PubLocalApp extends PubApp {
+  bool _activated = false;
+
+  _PubLocalApp(String appName) : super._(appName);
+
+  bool get isGlobal => false;
+
+  // TODO: For now, assume that local apps are installed?
```
seaneagan: I'm ok with the existing no-ops here, but to be correct I guess we'd have to to validate that the `packages` symlink has an entry for the package.  And for `activate` we'd have to run `Pub.get` first, which I don't think we should do.  

I wonder if it would be better to just throw an `UnsupportedError` for now here and in `activate` (and `update` if we still have it).  And then if we get a [pub activate](https://docs.google.com/document/d/1lvbzyivCejhTuHLG4WI3YVThbQqJhPUrwajwTC-sXWU/edit) in the future (and maybe a `pub list`), we can implement them using that.

seaneagan:
Done reviewing.

I also filed #166 for an outstanding comment from #153.

lib/grinder_tools.dart
```diff
@@ -310,50 +310,98 @@ class PubGlobal {
...
   void update() {
     Pub.global.activate(appName);
-    _installed = true;
+    _activated = true;
   }
 
-  String toString() => appName;
+  String run(List<String> arguments, {String script, String workingDirectory}) {
+    if (!_activated && !isActivated) activate();
+    return Pub.global.run(script == null ? appName : '${appName}:${script}',
```
seaneagan: Can you add a `script` param to `Pub.global.run` (see `Pub.run`) instead of passing in both as a single arg?

lib/grinder_tools.dart
```diff
@@ -310,50 +310,98 @@ class PubGlobal {
...
+
+  // TODO: For now, assume that local apps are installed?
+  bool get isActivated => true;
+
+  void activate() { }
+
+  void update() { }
+
+  String run(List<String> arguments, {String script, String workingDirectory}) {
+    return Pub.run(script == null ? appName : '${appName}:${script}',
```
seaneagan: Use `Pub.run`'s `script` param.

lib/grinder_tools.dart
devoncarew: changed to packageName

lib/grinder_tools.dart
devoncarew: I appreciate the brevity; right now I think I prefer the additional specificity though even if it makes the ctors a bit longer to reference.

lib/grinder_tools.dart
devoncarew: changed to throw a UnsupportedError

lib/grinder_tools.dart
devoncarew: done

lib/grinder_tools.dart
devoncarew: done

lib/grinder_tools.dart
devoncarew: I'm not sure I follow. Currently, the user doesn't have to explicitly activate a package if they want to use it, and we get caching about whether the package is installed or not (which takes shelling out to pub to determine).

devoncarew:
Thanks for the review! One question re: the `force: true` param for activation / running.

devoncarew:
(https://github.com/google/grinder.dart/commit/ee1dd15f54d9b7986bddc46ea4b57a91ee23bc5a)

seaneagan:
lgtm!

Issue #169, opened 2015-04-13
Incorrect command-line usage in README


seaneagan:
The README currently references `pub global run grind` which does not work.

The following all work:

`pub global activate grinder` followed by:

`grind`
`pub global run grinder`
`pub global run grinder:grind`
`pub global run grinder:grinder`

or bypass `pub global activate grinder` and just run:

`pub run grinder:grind`
`pub run grinder:grinder`
`pub run grinder` ([will work in dart 1.10](http://dartbug.com/22129))

IMO we should recommend both of the following:

* `pub run grinder:grind` (`pub run grinder` once dart 1.10 is released).  It's a single command, guarantees grinder version compatibility with the local package, and avoid global system state.
* `pub global activate` / `grind`.  It's really terse to run, and will allow for tab completion (#98).

devoncarew:
I like `pub global activate grinder` + `grind` for terseness. 

We should list `pub run grinder:grind` ( / `pub run grinder`) as well, and prefer the 2nd one.

seaneagan:
Fixed by #176.

PR #174, opened 2015-04-14
readme changes

devoncarew:
Fix https://github.com/google/grinder.dart/issues/169; @seaneagan

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -67,9 +67,10 @@ the command-line will run the default task if one is configured.
 
 or:
 
-    pub global run grind <args>
+    pub global run grinder <args>
 
-will run the `tool/grind.dart` script with the supplied arguments.
+will run the `tool/grind.dart` script with the supplied arguments. For for Dart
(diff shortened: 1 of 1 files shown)


seaneagan:
As in #169 I think we should mention `pub run grinder:grind` for pre-1.10 users, and not mention`pub global run ...` at all, since `pub global activate grinder` / `grind` should be preferred to that.  Also, it currently makes it look like `pub global activate grinder` is a required step, but that's only required if you want to use `grind`.  So "Installing" isn't always necessary... except for that the user needs to add grinder to their `dev_dependencies`, should we mention that?

devoncarew:
Let me call out two different ways of running it - globally activated and locally. I don't think we need to exhaustively list options though, just get people up and running quickly.

devoncarew:
updated -

seaneagan:
Thought it might be easiest to just create another pull request for comparison, see #176.  Also contains some other changes I've been wanting to do.

PR #183, opened 2015-04-15
Fix #150

### LINKED ISSUE #150: Deprecate TaskFunction
There's a note about it being deprecated in the doc-comments, but it should be marked `@Deprecated(...)` as well.

seaneagan:
@devoncarew

### DIFF
--- a/lib/grinder.dart
+++ b/lib/grinder.dart
@@ -24,6 +24,8 @@ import 'src/singleton.dart';
 /// Used to define a method body for a task. Note: a task's context is now
 /// available as a global variable ('context'). Your task functions should no
 /// longer be definied with a single `GrinderContext` parameter.
+@Deprecated('''Use a nullary function instead.  A task's context can now be
+accessed via the top-level `context` getter.''')
 typedef dynamic TaskFunction(GrinderContext context);
 
 /// Programmatically add a [task] to the global [Grinder] instance.
--- a/lib/grinder.dart
+++ b/lib/grinder.dart
@@ -28,6 +28,10 @@ import 'src/singleton.dart';
 accessed via the top-level `context` getter.''')
 typedef dynamic TaskFunction(GrinderContext context);
 
+// Temporary internal version of `TaskFunction`.
+// TODO: Remove this when removing the ability to use such functions.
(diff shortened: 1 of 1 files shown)


devoncarew:
Looks like the build system choked here: 

https://travis-ci.org/google/grinder.dart/builds/58621438#L184

because we're now using a deprecated typedef :)

I would either:

- relax our build to allow build infos (tuneup check --ignore-infos)
- in the failed line, instead of doing an `is` check for `TaskFunction`, check against a private typedef with the same sig

seaneagan:
Interesting, works in windows both locally and on appveyor.  `_TaskFunction` is definitely used.  Is this somehow a bug in the analyzer on linux?

devoncarew:
Do not know; this analyzes locally for me as well. I kicked off another copy of the build.

seaneagan:
So locally and on appveyor `pub global activate tuneup` uses analyzer 0.24.4, but on travis it is using 0.24.3.  Presumably it's a bug in analyzer 0.24.3 then.  Maybe the reason travis is using 0.24.3 is related to:

https://code.google.com/p/dart/issues/detail?id=23216#makechanges

seaneagan:
Should we also update grinder's analyzer constraint to ">=0.24.4 <0.25.0"?

seaneagan:
sorry, I mean tuneup's analyzer constraint

devoncarew:
https://github.com/google/tuneup.dart/commit/f9b1f4891a2214604b038482dc82bb5f4e2c030c

:)

devoncarew:
Travis is now pulling an old version of tuneup...

devoncarew:
To make progress, I would either commit as is and hope travis catches up, or change the build to use the `--ignore-infos` flag.

seaneagan:
ok, merging.  We can add `--ignore-infos` if it doesn't get resolved quickly.

seaneagan:
@devoncarew looks like the issue is fixed, can we kick off another travis build on master to fix the build status?  I either don't have access or can't find any "build now" button on the travis site.

devoncarew:
rebuilding now...

PR #191, opened 2015-04-20
make pub.global.list private for 0.7.0

devoncarew:
Make `PubGlobal.list` private for 0.7.0 - with plans to re-work the sig for 0.7.1.

@seaneagan


seaneagan:
lgtm, thanks!

PR #199, opened 2015-04-22
fixes to running w/o a task specified; cleanup help text

### LINKED ISSUE #119: improve the output of the dependency help text
Currently:

```
valid targets:
  [tests-build-web]
  [init]
  [analyze]
  [tests-web]
  [tests]
```

Since we'd moved to using annotations to define tasks, rather than programmatic setup in `main()`, its slightly harder to see the flow of tasks just by looking at the grind script. We should balance that by improving the output of the `-h` cli arg, to make the relationship between the tasks very visual and easy to parse.

### LINKED ISSUE #140: merge -h and -d options

devoncarew:
Fix #140, fix #119.

- fix an issue when running grinder w/ an invalid task name. It would print a long microqueue stack chain. It now prints `task 'foobar' doesn't exist` and exits with code 1.
- hide the `-d` option; aliased it to `-h`
- improved the help / deps output (see below)
- changed some tests to use a mock context when running
- added tests for some classes in the `src/utils` library

```
usage: dart tool/grind.dart <options> target1 target2 ...

valid options:
    --dart-sdk    set the location of the Dart SDK
    --version     print the version of grinder
-h, --help        show targets but don't build

targets:

[coverage]
  Gather and send coverage data.

[test]

[tests-build-web]

[analyze]

[tests-web]

[check-init]
  Check that the generated `init` grind script analyzes well.

[buildbot]
  @DefaultTask
  @Depends: [analyze] [test] [check-init] [coverage]
```

### DIFF
--- a/lib/src/cli.dart
+++ b/lib/src/cli.dart
@@ -39,10 +39,11 @@ Future handleArgs(List<String> args, {bool verifyProjectRoot: true}) {
         print('grinder is up to date!');
       }
     }).catchError((e) => null);
-  } else if (results['help']) {
-    printUsage(parser, grinder);
-  } else if (results['deps']) {
-    printDeps(grinder);
+  } else if (results['help'] || results['deps']) {
+    // TODO: Remove `deps` options post 0.7.0.
+    printUsageAndDeps(parser, grinder);
+  } else if (results.rest.isEmpty && !grinder.hasDefaultTask){
+    printUsageAndDeps(parser, grinder);
   } else {
     if (verifyProjectRoot) {
       // Verify that we're running from the project root.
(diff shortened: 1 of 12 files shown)


devoncarew:
@seaneagan

lib/src/cli.dart
```diff
@@ -74,28 +75,40 @@ ArgParser createArgsParser() {
...
-      if (grinder.defaultTask == t) buffer.write(' (default)');
-      if (t.description != null) buffer.write(' ${t.description}');
-      print(buffer.toString());
-    });
+    print(tasks.map((task) {
+      bool isDefault = grinder.defaultTask == task;
+      Iterable<GrinderTask> deps = grinder.getImmediateDependencies(task);
+
+      String str = '${task}\n';
+      if (isDefault) str += '  @DefaultTask\n';
```
seaneagan: I see how referencing the annotations in command-line help would be nice for devs familiar with the annotations.  But a dev who just wants to run the tasks, and doesn't know or want to know anything about how they are implemented, shouldn't have to, and might find it more difficult to decipher their meaning than if we used a more natural language approach.  We still have the programmatic `addTask` and `defaultTask = ...`, so the task could have been added that way instead.

So my preference would be something like:
```
[test]                  Run the tests.
[analyze]               Analyze the source code.
[buildbot] (default)    Tasks to run on the build bot.             
 ┣ [analyze]
 ┣ [test]
 ┣ [check-init]
 ┗ [coverage]
```

I like how you didn't show transitive dependencies, since I think that could get too busy / redundant.

seaneagan:
lgtm!

Have some ideas about how to make the output more visual, but that doesn't have to be in this PR.

seaneagan:
Can you add a note in the README that `grind -h` now shows the dependencies as well?

lib/src/cli.dart
devoncarew: The `@` callouts were mostly for to make it visually interesting; I changed it to `(defaults)` and `depends on` instead. I definitely want to make it pretty, as well as compact. I'll land as is and we can experiment with some other styles if necessary.

devoncarew:
> available tasks and their dependencies" 

sgmt, added.

devoncarew:
```
usage: dart tool/grind.dart <options> target1 target2 ...

valid options:
    --dart-sdk    set the location of the Dart SDK
    --version     print the version of grinder
-h, --help        show targets but don't build

targets:

[coverage]
  Gather and send coverage data.

[test]

[tests-build-web]

[analyze]

[tests-web]

[check-init]
  Check that the generated `init` grind script analyzes well.

[buildbot] (default)
  depends on: [analyze] [test] [check-init] [coverage]
```

PR #200, opened 2015-04-22
refactor file names, location, and contents

devoncarew:
- Move all the sdk related code from grinder_tools to `grinder_sdk.dart`
- rationalize our test dir layout a bit
- add some tests for things in grinder_tools

This is meant to be entirely w/o user impact, just some project hygiene. No changes to code (modulo some new tests).

@seaneagan


seaneagan:
Removing stuff from `grinder_tools.dart` is technically a breaking change.  Not sure if anyone is actually importing that library, but since it's just in `lib` rather than `lib/src`, it's fair game for them to.  I'm not sure we need any of the `lib/grinder_***.dart` libraries (including `grinder_sdk.dart` in this PR), since we already export them from `lib/grinder.dart`.  If we did keep them I think the `grinder_` prefix is redundant.

I personally like to have small libraries under `lib/src` which import each other as needed.  Then just export any public APIs from those in `lib/grinder.dart`.

devoncarew:
> Removing stuff from grinder_tools.dart is technically a breaking change.

Totally, but my guess is that most people are just importing only `grinder.dart`, esp. as that exports the other public libs. I'm comfortable with assuming the small change that we break a consumer to get the benefit of rev'ing the major semver fewer times (which has it's own cost downstream).

PR #201, opened 2015-04-22
coerce platform independent paths

devoncarew:
In the ctor of `FilePath`, coerce platform independent paths to platform dependent ones.

### DIFF
--- a/lib/grinder_files.dart
+++ b/lib/grinder_files.dart
@@ -117,6 +117,8 @@ class FilePath {
   /// Create a new [FilePath]. The [entityOrString] parameter can be a
   /// [FileSystemEntity] or a [String]. If a [String], this method converts the
   /// given path from a platform independent one to a platform dependent path.
+  /// This conversion will work for relative paths but wouldn't make sense to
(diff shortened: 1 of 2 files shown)


lib/grinder_files.dart
```diff
@@ -267,6 +268,8 @@ class FilePath {
 
   static String _coerce(arg) {
     if (arg is String) {
+      if (_sep != '/') arg = arg.replaceAll('/', _sep);
```
seaneagan: So this would convert e.g. `/foo/bar` to `\foo\bar` on windows.  Should that be something like `C:\foo\bar`?  Should either grinder (or grinder users) just be using the `path` package for this?

devoncarew: I'll add some more docs. This would have the same behavior as the `getFile()` / `getDir()` methods late in the file. They'll work for relative paths, but it would not make sense for absolute ones. Those will likely be machine dependent anyway. (and auto-converting `/foo/bar` to `C:\foo\bar` would be difficult - we wouldn't know what drive to cvt to)

PR #202, opened 2015-04-22
add dartfmt support

### LINKED ISSUE #168: Request: default task: format
I'd love it if grinder had a default talk for 'format' which ran dart_style's dartfmt over my project.

Thanks!

devoncarew:
Fix #168.

@seaneagan 

Write a wrapper around the `dartfmt` SDK tool. This is scoped to just exposing what's in the SDK .The user still has the option of using a `PubApp` wrapper around the `dart_style` tool.

(the `Analyzer` and `Dart2js` classes are unchanged; they just changed locations within the file)

### DIFF
--- a/lib/grinder_sdk.dart
+++ b/lib/grinder_sdk.dart
@@ -33,13 +33,9 @@ Directory getSdkDir([List<String> cliArgs]) => cli_util.getSdkDir(cliArgs);
 
 File get dartVM => joinFile(sdkDir, ['bin', _sdkBin('dart')]);
 
-/// Utility tasks for for getting information about the Dart SDK and invoking
-/// dart applications.
+/// Utility tasks for for getting information about the Dart VM and for running
+/// Dart applications.
 class Dart {
-  /// Return the path to the current Dart SDK. This will return `null` if we are
-  /// unable to locate the Dart SDK.
-  static Directory get location => sdkDir;
-
   /// Run a dart [script] using [run_lib.run].
   ///
   /// Returns the stdout.
@@ -77,6 +73,12 @@ class Dart {
     return Platform.version.substring(0, Platform.version.indexOf(' '));
   }
 }
+
+//class DartSdk {
+//  /// Return the path to the current Dart SDK. This will return `null` if we are
+//  /// unable to locate the Dart SDK.
+//  static Directory get location => sdkDir;
+//}
 
 /**
  * Utility tasks for executing pub commands.
--- a/lib/grinder_sdk.dart
+++ b/lib/grinder_sdk.dart
@@ -269,7 +269,8 @@ class Analyzer {
       _sdkBin('dartanalyzer'), quiet: quiet, arguments: ['--version'])).version;
 }
 
-/// Utility class for invoking `dartfmt` from the SDK.
+/// Utility class for invoking `dartfmt` from the SDK. This wrapper requires
+/// the `dartfmt` from SDK 1.9 and greater.
 class DartFmt {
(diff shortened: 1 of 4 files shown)


lib/grinder_sdk.dart
```diff
@@ -33,6 +33,51 @@ Directory getSdkDir([List<String> cliArgs]) => cli_util.getSdkDir(cliArgs);
 
 File get dartVM => joinFile(sdkDir, ['bin', _sdkBin('dart')]);
 
+/// Utility tasks for for getting information about the Dart SDK and invoking
+/// dart applications.
+class Dart {
```
devoncarew: Should this class be called `Dart` or `DartSdk`?

devoncarew: DartVM?

seaneagan: I chose `Dart` since it wraps the `dart` executable, similar to how `Pub` wraps `pub` and `Dart2js` wraps `dart2js`.  However we do have `Analyzer` for `dartanalyzer, so we could do `Dart2js` -> `Compiler` and `Dart` -> `Vm`, but I kind of think the `Dart` prefix is nice.

seaneagan: Hmm, it might be interesting to namespace all the sdk stuff, so:

```dart
class Sdk {

  /// Defaults the path to the current SDK.
  Sdk([String path]);
  
  static Directory get dir;

  final Vm vm;
  final Compiler compiler;
  final Analyzer analyzer;
  final Formatter formatter;
}

final Sdk sdk = new Sdk();
```

```dart
print('Sdk dir: ${sdk.dir}');
sdk.vm.run(...);
sdk.compiler.compile(...);
sdk.analyzer.analyze(...);
sdk.formatter.format(...);
```

lib/grinder_sdk.dart
```diff
@@ -146,6 +191,103 @@ class Pub {
...
+
+  static String version({bool quiet: false}) => _AppVersion.parse(run_lib.run(
+      _sdkBin('dartanalyzer'), quiet: quiet, arguments: ['--version'])).version;
+}
+
+/// Utility class for invoking `dartfmt` from the SDK.
+class DartFmt {
+  /// Run the `dartfmt` command with the `--overwrite` option. Format any files
+  /// in place.
+  static void format(fileOrPath) {
```
seaneagan: I think we'd be better off moving everything over to constructors and instance methods instead of classes as containers for static methods.  That would allow them to be:

* configured via constructors
* passed around as instances
* mocked out for testing

So:

```dart
class Pub {
  Pub();
  get(...);
  //...
}

final Pub pub = new Pub();

class Formatter {
  Formatter.sdk(...);
  Formatter([VersionConstraint constraint]);
  //...
}
```

```dart
pub.global.run(...);
sdk.formatter.format(...); // As suggested in another review comment.
new Formatter().format(...); // Uses pub.
```

seaneagan: It's kind of weird the `Pub` uses static members, but `PubGlobal` uses instance members.  This would fix that too.

seaneagan:
Filed my review comments as #204, let's discuss there.

lib/grinder_sdk.dart
devoncarew: I left this as `Dart`, and removed the `location` static field. That would apply to a wrapper around the dartsdk, but not one around the dartvm.

lib/grinder_sdk.dart
devoncarew: I'll comment there; I generally like the idea but am worried about the cost of more breaking changes.

PR #209, opened 2015-04-24
rename some global variables

### LINKED ISSUE #206: Refactor file path shortcuts
Do we need the following?

`WEB_DIR`, `LIB_DIR`, `BUILD_DIR`, `BIN_DIR`, `dartVM`.

The names all violate the style guide, which would suggest:

`webDir`, `libDir`, `buildDir`, `binDir`, `dartVm`

The dirs could make sense as part of a class which represents a pub package:

```
class Package {
  Directory get lib => new Directory('lib');
  Directory get bin => new Directory('bin');
  Directory get test => new Directory('test');
  Directory get web => new Directory('web');
  Directory get build => new Directory('build');
  // Anything else here?
}

Package package => new Package();
```

`dartVm` could make sense as part of #204:

`vm.path`

devoncarew:
Fix #206.

@seaneagan

### DIFF
--- a/lib/grinder_tools.dart
+++ b/lib/grinder_tools.dart
@@ -21,10 +21,7 @@ export 'src/run.dart';
 
 final Directory binDir = new Directory('bin');
 final Directory buildDir = new Directory('build');
-final Directory exampleDir = new Directory('example');
 final Directory libDir = new Directory('lib');
-final Directory testDir = new Directory('test');
-final Directory toolDir = new Directory('tool');
 final Directory webDir = new Directory('web');
 
 @Deprecated('See binDir') Directory get BIN_DIR => binDir;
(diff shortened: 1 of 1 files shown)


seaneagan:
lgtm.

It'd be cleaner and better for code completion if we could namespace them somehow, like a separate `lib/package_dirs.dart` or a `Package` or `Dirs` class with statics, and remove the `Dir` suffix, so e.g. `bin`, `lib` members.

devoncarew:
Hmm, yeah. I'd like to keep them as short as possible, but I see the issue with polluting the namespace. Perhaps this is an argument for just exposing the most used ones?

seaneagan:
sgtm.  I'll let you decide which ones that is.  The original 4 you had might be reasonable, but `bin` may not be that common now that we have `pub run`, and I guess `lib` might be useful for codegen?

devoncarew:
I think I'll roll back to just the 4 that were already available.

PR #213, opened 2015-04-29
Use unscripted.  Fix #98.  Improve help output.

### LINKED ISSUE #98: Tab completion for tasks
It would be really cool to be able to tab-complete task names on the command line.  This would be super easy to do using [unscripted](https://pub.dartlang.org/packages/unscripted).

seaneagan:
@devoncarew

### DIFF
--- a/lib/grinder.dart
+++ b/lib/grinder.dart
@@ -317,10 +317,9 @@ class Grinder {
     if (targets.isEmpty) {
       if (defaultTask != null) {
         targets = [defaultTask.name];
-      } else {
-        if (!dontRun) log('no tasks specified, and no default task defined');
+      } else if (!dontRun) {
+        throw new GrinderException('Tried to call non-existent default task.');
       }
-      if (!dontRun) log('run `grinder -h` for help and a list of valid tasks\n');
       if (targets.isEmpty) return new Future.value();
     }
 
--- a/lib/src/cli.dart
+++ b/lib/src/cli.dart
@@ -5,11 +5,12 @@ library grinder.src.cli;
 
 import 'dart:async';
 import 'dart:convert' show JSON, UTF8;
-import 'dart:io';
+import 'dart:math';
 
-import 'package:args/args.dart';
+import 'package:ansicolor/ansicolor.dart';
+import 'package:unscripted/unscripted.dart';
 
-import 'singleton.dart';
+import 'singleton.dart' as singleton;
 import 'utils.dart';
 import '../grinder.dart';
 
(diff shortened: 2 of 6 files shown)


seaneagan:
CI failure is https://github.com/google/tuneup.dart/issues/30

lib/grinder.dart
```diff
@@ -317,10 +317,9 @@ class Grinder {
     if (targets.isEmpty) {
       if (defaultTask != null) {
         targets = [defaultTask.name];
-      } else {
-        if (!dontRun) log('no tasks specified, and no default task defined');
+      } else if (!dontRun) {
+        throw new GrinderException('Tried to call non-existent default task.');
```
devoncarew: It looks like this change will mean that running grinder w/o a task specified will print `Tried to call non-existent default task` instead of `run 'grinder -h' for help and a list of valid tasks`. I think we're better off w/ this not being an error condition -

seaneagan: I moved the error to the cli-specific code:

https://github.com/google/grinder.dart/pull/213/files#diff-48f50be87f4906250da81d6ab77b472aR60

So the cli will catch this before `Grinder.start` is even called.  But I still kept an exception here to cover programmatic usage of `Grinder.start`.

devoncarew:
To confirm, the --help output now looks like:
```
[a]      1
[b]      2
[ab]     (depends on [a] [b])
[abc]   123
          (depends on [ab])
```

And re: the build failure, can you disable the call to tuneup in the grinder script, so we can keep test coverage w/ this PR?

lib/grinder.dart
devoncarew: Ah, ok.

seaneagan:
Here's the complete help output:

grind_help_output

seaneagan:
re: the travis failure, looks like it doesn't occur in Windows since it's working for me locally and on AppVeyor.  I assume you are on Mac or Linux, can you see if you get it locally?

devoncarew:
I can see the failure locally - the exception is showing up when running the dartanalyzer as well.

devoncarew:
https://code.google.com/p/dart/issues/detail?id=23354

lib/src/cli.dart
```diff
@@ -18,41 +19,55 @@ const String APP_VERSION = '0.7.0';
...
   _args = args == null ? [] : args;
+  _verifyProjectRoot = verifyProjectRoot;
+  return script.execute(grinderArgs());
+}
 
-  ArgParser parser = createArgsParser();
-  ArgResults results = parser.parse(grinderArgs());
-
-  if (results['version']) {
+@Command(help: 'Dart workflows, automated.', plugins: const [const Completion()])
```
stereotype441: This is the line that's causing the analyzer to crash (see http://dartbug.com/23354).  If you don't want to wait for the bug fix, you should be able to work around the problem with no change in semantics by moving the instance creation expression to a separate constant, e.g.:

    const _completion = const Completion();
    @Command(help: 'Dart workflows, automated.', plugins: const [_completion])

seaneagan: @stereotype441 yep, that works until the fix is released, thanks!

devoncarew:
lgtm!

PR #216, opened 2015-05-02
accept a list of VM args when running the Dart task

sethladd:


devoncarew:
Looks like we're not `unittest` 0.12.0 happy. @sethladd, can you peg the dep at ^0.11.0 in this CL, and we'll clean up our tests in a separate one?

sethladd:
weird. tests passed on my machine.

sethladd:
I just ran my tests via `dart test/all.dart` so maybe it's an issue with grinder and the new test runner.

sethladd:
Look good, please review again. Thanks!

devoncarew:
Lgtm

PR #220, opened 2015-05-06
use the webkit_inspection_protocol package

devoncarew:
Move the `_wip.dart` library into a new `webkit_inspection_protocol` package; reference that package from grinder.

@seaneagan


seaneagan:
lgtm, but isn't it called the [remote debugging protocol](https://developer.chrome.com/devtools/docs/debugger-protocol) now?

devoncarew:
Well, WIP and the remote debugging protocol are still the same thing. If they do diverge, the `webkit_inspection_protocol.dart` package will track the remote debugging protocol. In which case I may or may not rename the repo -

zoechi:
Interesting. Curious what you plan to use it for.

PR #221, opened 2015-05-08
add `pub downgrade`.

zoechi:
I want to run my tests twice to check whether the lower bounds of my dependencies are still appropriate.

```dart
_test() {
  Pub.downgrade();
  new PubApp.local('test').run(['-rexpanded']);
  Pub.upgrade();
  new PubApp.local('test').run(['-rexpanded']);
}
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,3 +10,4 @@ W. Brian Gourlie <bgourlie@gmail.com>
 Kenneth Endfinger <kaendfinger@gmail.com>
 Sean Eagan <seaneagan1@gmail.com>
 Guillermo López-Anglada <guillermo.lopez@outlook.com>
+Günter Zöchbauer <guenter@gzoechbauer.com>
--- a/lib/grinder_sdk.dart
+++ b/lib/grinder_sdk.dart
@@ -138,6 +138,21 @@ class Pub {
   }
 
   /**
+   * Run `pub downgrade` on the current project.
+   */
+  static void downgrade({String workingDirectory}) {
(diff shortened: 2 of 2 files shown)


devoncarew:
lgtm!

Can you also add yourself to the AUTHORS file?

zoechi:
done

devoncarew:
Thanks! Merging...

seaneagan:
@devoncarew should we be updating the `authors` field in the pubspec as well, so it gets reflected on pub.dartlang.org?

zoechi:
Does it make sense to have two places for that?
I would either leave `pubspec.yaml` as it is and maintain the `AUTHORS` file or remove the authors file completely and maintain only the `pubspec.yaml` file.
A 3rd way is to add a grinder task that updates `AUTHORS` from `pubspec.yaml` ;-)

devoncarew:
They're two separate things. The AUTHORS file is necessary to accept contributions and make sure that the CLA's been signed, and copyright assigned. We can't really mess with the AUTHORS file, it's part of Google's IP process.

I'm not sure the pubspec field is useful for a lot -

devoncarew:
Although this reminds me, we should be updating the changelog :)

zoechi:
done (new PR)

seaneagan:
The pubspec field is what drives the author display in the [pub package page](https://pub.dartlang.org/packages/grinder).  Not super important though since github already [provides](https://github.com/google/grinder.dart/graphs/contributors) an automatically updated view of authors.

Issue #222, opened 2015-05-08
Checked mode error in grind script


zoechi:
I tried to debug my grinder script but when I don't start it with `grind task` but instead with
`dart -c tool/grind.dart task`

I get 

```
Unhandled exception:
Uncaught Error: Failed assertion: boolean expression must not be null
Stack Trace:
#0      cli (package:grinder/src/cli.dart:47:7)
#1      Function._apply (dart:core-patch/function_patch.dart:7)
#2      Function.apply (dart:core-patch/function_patch.dart:28)
#3      FunctionScript._getTopCommandResult (package:unscripted/src/script_impl.dart:154:67)
#4      DeclarationScript._handleResults (package:unscripted/src/script_impl.dart:104:21)
#5      ScriptImpl.execute.<anonymous closure>.<anonymous closure> (package:unscripted/src/script_impl.dart:49:11)
#6      _RootZone.runUnary (dart:async/zone.dart:1166)
#7      _Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:494)
#8      _Future._propagateToListeners (dart:async/future_impl.dart:577)
#9      _Future._completeWithValue (dart:async/future_impl.dart:368)
#10     _Future._asyncComplete.<anonymous closure> (dart:async/future_impl.dart:422)
#11     _microtaskLoop (dart:async/schedule_microtask.dart:43)
#12     _microtaskLoopEntry (dart:async/schedule_microtask.dart:52)
#13     _runPendingImmediateCallback (dart:isolate-patch/isolate_patch.dart:96)
#14     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:149)

#0      _rootHandleUncaughtError.<anonymous closure> (dart:async/zone.dart:895)
#1      _microtaskLoop (dart:async/schedule_microtask.dart:43)
#2      _microtaskLoopEntry (dart:async/schedule_microtask.dart:52)
#3      _runPendingImmediateCallback (dart:isolate-patch/isolate_patch.dart:96)
#4      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:149)

Process finished with exit code 255
```

seaneagan:
[...]

PR #223, opened 2015-05-08
run grinder in checked mode on the bot

devoncarew:
Related to #222, run grinder in checked mode on the bots (and fix one checked mode error).

### DIFF
--- a/.project
+++ b/.project
@@ -10,9 +10,15 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
+			<name>org.dartlang.tools.dartBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
 	</buildSpec>
 	<natures>
 		<nature>com.google.dart.tools.core.dartNature</nature>
+		<nature>org.dartlang.tools.dartNature</nature>
 	</natures>
 	<filteredResources>
 		<filter>
--- a/.project
+++ b/.project
@@ -10,15 +10,9 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
-		<buildCommand>
-			<name>org.dartlang.tools.dartBuilder</name>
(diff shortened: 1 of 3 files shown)


lib/src/cli.dart
```diff
@@ -35,12 +35,13 @@ cli(
     @Rest(help: getTaskHelp, allowed: allowedTasks)
     List<String> tasks,
     {
-     @Flag(help: 'Print the version of grinder.')
+     @Flag(help: 'Print the version of grinder.', defaultsTo: false)
      bool version,
```
seaneagan: Nit: You can just do `bool version: false`.  I like to define as much of the interface using dart-native features as possible.

.travis.yml
```diff
@@ -1,3 +1,3 @@
 language: dart
-script: pub run grinder:grinder
+script: dart -c tool/grind.dart buildbot
```
seaneagan: Nit: Can we remove `buildbot` to dogfood the default task handling?

lib/src/cli.dart
devoncarew: sgtm. I like this better as well; I got a bit confused by the optional params and param annotations -

.travis.yml
devoncarew: done

devoncarew:
updated -

.project
```diff
@@ -10,9 +10,15 @@
 			<arguments>
```
seaneagan: Should this file be .gitignored?

seaneagan: I guess this file could probably be .gitignore'd (I thought it was already in github's default .gitignore for Dart), not that it hurts anything having it.

devoncarew: Hmm, in any case I didn't mean to have it in this PR -

seaneagan:
lgtm

PR #227, opened 2015-05-10
support to pass environment arguments to `Process.start`/`Process.run…

zoechi:
…` for all runXxx commands

I run into situations where I missed this several times already.
Currently it's for running tests where I need to pass Docker connection information to the tests.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -4,7 +4,9 @@ version: 0.7.0
 description: Dart workflows, automated.
 
 homepage: https://github.com/google/grinder.dart
-author: Devon Carew <devoncarew@google.com>
+authors:
+- Devon Carew <devoncarew@google.com>
+- Sean Eagan <seaneagan1@gmail.com>
 
 environment:
   sdk: '>=1.0.0 <2.0.0'


seaneagan:
We probably want `bool includeParentEnvironment: true` also, and we're missing some others as well: `bool runInShell: false, Encoding stdoutEncoding: SYSTEM_ENCODING, Encoding stderrEncoding: SYSTEM_ENCODING.

What about packaging up process running args in an object:

```
class RunArgs {
  final List<String> arguments;
  final String workingDirectory;
  final Map<String, String> environment;
  final bool includeParentEnvironment;
  final bool runInShell;
  final Encoding stdoutEncoding;
  final Encoding stderrEncoding;

  RunArgs({
      this.arguments: const [],
      this.workingDirectory,
      this.environment,
      this.includeParentEnvironment: true,
      this.runInShell: false,
      this.stdoutEncoding: SYSTEM_ENCODING,
      this.stderrEncoding: SYSTEM_ENCODING});
}

String run(String executable, RunArgs runArgs);
Future<String> runAsync(String executable, RunArgs runArgs);
```

Other methods receive a `RunArgs` as well, but can tweak it as needed before forwarding to `run` and `runAsync`.  For example `Pub.global.run('foo', script: 'bar')` would add `global run foo:bar` to the arguments.

It'd be a breaking change, but we could keep the old deprecated `runProcess` and `runProcessAsync` a little longer so that we only break users ot `run` and `runAsync` introduced in 0.7.0.

zoechi:
I wasn't sure what the intention omitting these arguments. 
I was thinking along the same lines.
It would be nice to have all arguments available `Process.start/run` can handle and I also think it would be better to have such an arguments class instead forwarding a lot of arguments several times in several places.

zoechi:
I'm wondering if we should leave `arguments` as it is. It's the most common and maybe cumbersome to create an instance each time and the hardest breaking change. 
I assume `workingDirectory` is much less used (but still a breaking change).
I tend to move `arguments` and `workingDirectory` to RunArgs as suggested by @seaneagan. What's your opinion?

zoechi:
What about other commands that use `Process.run`/`Process.start` like `Pub.get`? Should there also `RunArgs` be used? 
Might be possible that here additional arguments are useful eventually. 
For example it doesn't support `arguments` currently but `pub get` **does** support additional arguments.

devoncarew:
@zoechi, thanks for the PR! I like the formatting changes, but in the future can you send them as a separate PR or as a different commit in the same PR? It'll make it easier to tease apart the formatting changes and the functional ones.

Agree w/ @seaneagan that the methods are getting a lot of params at this point. Exposing the env args makes sense; definitely also arguments and workingDirectory. I would also like to avoid adding more deprecated methods and introducing more breaking changes. 

Will we ever need the less used options (includeParentEnvironment, runInShell, stdoutEncoding, stderrEncoding)? Could we just expose them in the runlib's `run()` method?

seaneagan:
I'm in favor of:

* adding a `RunOptions` (or similar) class
* deprecating `workingDirectory` (and `quiet`?  it's grinder-specific, but would be nice to have everywhere) named arguments in favor of `RunOptions` properties.
* adding `RunOptions options` named args to (most) process running APIs.  Some things like `Pub.version()` may not need it.
* Ideally it'd be `List<String> | RunOptions options` named parameters, but since we don't yet have union types, keeping separate `arguments` parameters as suggested by @zoechi sounds good.

I think it'd be easier to grok both the impl and interface, learning about `RunOptions` once, and then knowing how to use it in any process running API which all have very simple (and consistent) signatures.

zoechi:
Sorry for the inconvenience. I'll try to follow this guideline.

I wasn't able yet to figure out what `includeParentEnvironment` actually does. I think I know what `runInShell` does. I wasn't aware of `xxxEncodind` because I usually use `start` instead of `run`. 

I mostly implemented the RunArgs approach but wasn't able to finish today. I'll make a PR tomorrow so you can have a look and to have some material for further discussion. 

Adding the `arguments` to the `RunArgs` makes some methods more complicated. I think we should leave this out. I would keep the rest though.

zoechi:
I would do exactly as @seaneagan wrote.

I'm not sure how to cope with the different set of arguments for `Process.run` and `Process.start`
I could create `StartOptions` and `RunOptions` with their different set of arguments but I think this is not optimal because the names in Grinder don't use `start`/`run` depending on which `Process.xxx` gets called in the end.
What about using `ProcessOptions` with the common set of arguments and derived `ProcessRunOptions`/`ProcessStartOptions` with their additional arguments.
The `processOptions` argument in the methods would be typed `ProcessOptions` but `RunOptions`/`StartOptions` could be supported as well when these less common options are needed.
This way the options would be the same for the common cases but with the price of less discoverability for the additional arguments.

```dart
class ProcessOptions {
  final List<String> arguments;
  final bool quiet;
  final String workingDirectory;
  final Map<String, String> environment;
  final bool includeParentEnvironment;
  final bool runInShell;
// left out constructors for brevity
}

class ProcessRunOptions extends ProcessOptions {
  final Encoding stdoutEncoding;
  final Encoding stderrEncoding;
}

class ProcessStartOptions extends ProcessOptions {
  ProcessStartMode mode;
}

/// Document here that RunOptions can be used instead of ProcessOptions
String run(String executable, {ProcessOptions options});
/// Document here that StartOptions can be used instead of ProcessOptions
Future<String> runAsync(String executable, {ProcessOptions options});
```
What do you think?

zoechi:
A better fit for Grinder might be to use `RunOptions`/`AsyncRunOptions` and then using this types also in the method signature.

zoechi:
Is it ok to use `dartformat`?

PR #228, opened 2015-05-12
Introduce `RunOptions`/`RunAsyncOptions` to support all `Process.run`…

zoechi:
…/`Process.start` parameters

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,8 +2,8 @@
 
 ## UNRELEASED
 - Introduce `RunOptions` to support all `Process.run`/`Process.start` 
-  parameters. This is a breaking change because it replaces the `workingDirectory` 
-  parameter in several methods.
+  parameters. The `workingDirectory` parameter is now deprecated in all methods, 
+  use `RunOptions.workingDirectory` instead.
 
 ## 0.7.1
 - `Dart.run` now takes an optional `vmArgs`, a list of arguments passed to the Dart VM.
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,7 @@
 # grinder.dart changes
 
 ## UNRELEASED
-- Introduce `RunOptions`/`RunAsyncOptions` to support all `Process.run`/`Process.start` 
+- Introduce `RunOptions` to support all `Process.run`/`Process.start` 
   parameters. This is a breaking change because it replaces the `workingDirectory` 
   parameter in several methods.
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,13 +1,11 @@
 # grinder.dart changes
 
 ## UNRELEASED
+- `Dart.run` now takes an optional `vmArgs`, a list of arguments passed to the Dart VM.
+- Added `downgrade` method do `Pub`.
 - Introduce `RunOptions` to support all `Process.run`/`Process.start` 
   parameters. The `workingDirectory` parameter is now deprecated in all methods, 
   use `RunOptions.workingDirectory` instead.
-
-## 0.7.1
-- `Dart.run` now takes an optional `vmArgs`, a list of arguments passed to the Dart VM.
-- Added `downgrade` method do `Pub`.
 
 ## 0.7.0
 - Big changes! Task definititions are now primarily annotation based; see the
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,10 @@
 # grinder.dart changes
 
+## UNRELEASED
+- Introduce `RunOptions`/`RunAsyncOptions` to support all `Process.run`/`Process.start` 
+  parameters. This is a breaking change because it replaces the `workingDirectory` 
+  parameter in several methods.
+
 ## 0.7.1
 - `Dart.run` now takes an optional `vmArgs`, a list of arguments passed to the Dart VM.
 - Added `downgrade` method do `Pub`.
--- a/lib/grinder_sdk.dart
+++ b/lib/grinder_sdk.dart
@@ -12,6 +12,7 @@ import 'package:which/which.dart';
 
 import 'grinder.dart';
 import 'src/run.dart' as run_lib;
+import 'src/run_utils.dart';
 
 bool _sdkOnPath;
 
--- a/lib/grinder_sdk.dart
+++ b/lib/grinder_sdk.dart
@@ -110,7 +110,7 @@ class Pub {
    * even if the pubspec.lock file is up-to-date with respect to the
    * pubspec.yaml file.
    */
-  static Future getAsync({bool force: false, RunAsyncOptions runOptions}) {
+  static Future getAsync({bool force: false, RunOptions runOptions}) {
     FileSet pubspec = new FileSet.fromFile(new File('pubspec.yaml'));
     FileSet publock = new FileSet.fromFile(new File('pubspec.lock'));
 
@@ -132,7 +132,7 @@ class Pub {
   /**
    * Run `pub upgrade` on the current project.
    */
-  static Future upgradeAsync({RunAsyncOptions runOptions}) {
+  static Future upgradeAsync({RunOptions runOptions}) {
     return run_lib.runAsync(_sdkBin('pub'), arguments: ['upgrade'],
         runOptions: runOptions).then((_) => null);
   }
@@ -147,7 +147,7 @@ class Pub {
   /**
    * Run `pub downgrade` on the current project.
    */
-  static Future downgradeAsync({RunAsyncOptions runOptions}) {
+  static Future downgradeAsync({RunOptions runOptions}) {
     return run_lib.runAsync(_sdkBin('pub'), arguments: ['downgrade'],
         runOptions: runOptions).then((_) => null);
   }
@@ -179,7 +179,7 @@ class Pub {
   static Future buildAsync({
       String mode,
       List<String> directories,
-      RunAsyncOptions runOptions,
+      RunOptions runOptions,
       String outputDirectory}) {
     List args = ['build'];
     if (mode != null) args.add('--mode=${mode}');
--- a/lib/grinder_sdk.dart
+++ b/lib/grinder_sdk.dart
@@ -47,7 +47,7 @@ class Dart {
       @Deprecated('Use RunOptions.workingDirectory instead.')
       String workingDirectory, int vmNewGenHeapMB,
        int vmOldGenHeapMB, List<String> vmArgs : const []}) {
-    runOptions = mergeRunOptions(workingDirectory, runOptions);
+    runOptions = mergeWorkingDirectory(workingDirectory, runOptions);
     List<String> args = [];
 
     args.addAll(vmArgs);
@@ -101,7 +101,7 @@ class Pub {
   static void get({bool force: false, RunOptions runOptions,
       @Deprecated('Use RunOptions.workingDirectory instead.')
       String workingDirectory}) {
-    runOptions = mergeRunOptions(workingDirectory, runOptions);
+    runOptions = mergeWorkingDirectory(workingDirectory, runOptions);
     FileSet pubspec = new FileSet.fromFile(new File('pubspec.yaml'));
     FileSet publock = new FileSet.fromFile(new File('pubspec.lock'));
 
@@ -118,7 +118,7 @@ class Pub {
   static Future getAsync({bool force: false, RunOptions runOptions,
       @Deprecated('Use RunOptions.workingDirectory instead.')
       String workingDirectory}) {
-    runOptions = mergeRunOptions(workingDirectory, runOptions);
+    runOptions = mergeWorkingDirectory(workingDirectory, runOptions);
     FileSet pubspec = new FileSet.fromFile(new File('pubspec.yaml'));
     FileSet publock = new FileSet.fromFile(new File('pubspec.lock'));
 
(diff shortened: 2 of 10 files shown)


zoechi:
See also #227

lib/grinder_sdk.dart
```diff
@@ -44,7 +44,7 @@ class Dart {
   /// Returns the stdout.
   static String run(String script,
       {List<String> arguments : const [], bool quiet: false,
-       String packageRoot, String workingDirectory, int vmNewGenHeapMB,
```
seaneagan: Do we want to keep `workingDirectory` as a deprecated parameter for now, and use it when provided, and throw if both `workingDirectory` and `runOptions` are non-null?

lib/grinder_sdk.dart
```diff
@@ -44,7 +44,7 @@ class Dart {
...
+       String packageRoot, RunOptions runOptions, int vmNewGenHeapMB,
```
seaneagan: I think just `options` would be clear enough when calling the `run` and `runAsync` method.  Question is whether it would be clear when calling e.g. `Dartfmt.format(options: new RunOptions(...))`.  So I'd say stick with `runOptions` everywhere to be consistent.

lib/grinder_sdk.dart
zoechi: There is no way to deprecate a parameter?

lib/grinder_sdk.dart
zoechi: I had the same thoughts, tried `options` first but had the impression this could be confusing.

lib/src/run.dart
```diff
@@ -144,3 +152,56 @@ stderr:
...
+  final Encoding stderrEncoding;
+
+  RunOptions({this.workingDirectory, this.environment,
+      this.includeParentEnvironment: true, this.runInShell: false,
+      this.stdoutEncoding: SYSTEM_ENCODING,
+      this.stderrEncoding: SYSTEM_ENCODING});
+
+  /// Create a clone when it's necessary to modify the passed runOptions to
+  /// avoid modifying the argument.
+  RunOptions clone() {
```
seaneagan: Since all the fields are final, I'm not sure how you would update the returned clone with the changed values.  The pattern I've seen for this that I like is a method which returns an object with the same fields as an existing instance, except for explicit overrides.  It could either be a constructor:

(using ?? for brevity)

```
  RunOptions.from(RunOptions other, {String workingDirectory, Map<String, String> environment,
      bool includeParentEnvironment, bool runInShell,
      Encoding stdoutEncoding,
      Encoding stderrEncoding}) : this(
      workingDirectory: workingDirectory ?? other.workingDirectory,
      environment: environment ?? other.environment,
      includeParentEnvironment: includeParentEnvironment ?? other.includeParentEnvironment,
      stdoutEncoding: stdoutEncoding ?? other.stdoutEncoding,
      stderrEncoding: stderrEncoding ?? other.stderrEncoding
);
```

or an instance method:

```
  RunOptions with({String workingDirectory, Map<String, String> environment,
      bool includeParentEnvironment, bool runInShell,
      Encoding stdoutEncoding,
      Encoding stderrEncoding}) => new RunOptions(
      workingDirectory: workingDirectory ?? this.workingDirectory,
      environment: environment ?? this.environment,
      includeParentEnvironment: includeParentEnvironment ?? this.includeParentEnvironment,
      stdoutEncoding: stdoutEncoding ?? this.stdoutEncoding,
      stderrEncoding: stderrEncoding ?? this.stderrEncoding
);
```

zoechi: Great you noticed. I introduced this for `arguments` in the previous implementation and decided to keep it, seems I removed `final` only for arguments and forgot the others.

zoechi: I preferred the `clone` solution because the clone code is simpler and because I don't need `new`
`runOptions.clone()..environment['xxx']='yyy'`   
instead of   
`new RunOptions.from(runOptions, environment: new Map.from(runOptions.environment)..['xxx']='yyy')`

zoechi:
I removed final but it's now part of #229.

lib/src/run.dart
```diff
@@ -144,3 +152,56 @@ stderr:
...
+        includeParentEnvironment: includeParentEnvironment,
+        runInShell: runInShell,
+        stdoutEncoding: stdoutEncoding,
+        stderrEncoding: stderrEncoding);
+  }
+}
+
+/// Arguments passed to [Process.start] .
+/// See [Process.start] for more details.
+class RunAsyncOptions {
```
seaneagan: I don't think `RunAsyncOptions` is needed.

Firstly, `runAsync` is just like `run`, except for the output is live streamed to grinder's logging as the process is running.  So `stdoutEncoding` and `stderrEncoding` would be useful to `runAsync` (and its callers) as well to decode output before logging.

Secondly, the functionality of `runAsync` to throw an exception on non-zero exit code would not be available anyway when starting a process in `DETACHED` or `DETACHED_WITH_STDIO` mode.  And the functionality to automatically `log` the output would not be available in `DETACHED` mode.  So I don't think exposing the `mode` will be very useful.

So combining those two, both `run` and `runAsync` (and their callers) would benefit from exactly the fields you currently have defined in `RunOptions`.

Agree?

zoechi:
Sounds reasonable. I didn't use these parameters yet and didn't investigate what they are for. Getting rid of one of those classes would be great. I'll have a closer look tomorrow, it's time for bed here.

lib/grinder_sdk.dart
zoechi: http://dartbug.com/22975
`@deprecated` on an optional argument doesn't produce a hint/warning.

zoechi:
I removed `RunAsyncOptions`. The removed `finals` are now also removed in this PR (sorry, Git still causes me some troubles)
Keeping workingDirectory with a deprecated annotation IMHO doesn't provide any value when there is no warning, so I left it removed.

lib/src/run.dart
seaneagan: It'd be really nice to keep the fields final.  What about:

(note the Map copy)

```
  RunOptions withOptions({String workingDirectory, Map<String, String> environment,
      bool includeParentEnvironment, bool runInShell,
      Encoding stdoutEncoding,
      Encoding stderrEncoding}) => new RunOptions(
      workingDirectory: workingDirectory ?? this.workingDirectory,
      environment: environment ?? new Map.from(this.environment),
      includeParentEnvironment: includeParentEnvironment ?? this.includeParentEnvironment,
      stdoutEncoding: stdoutEncoding ?? this.stdoutEncoding,
      stderrEncoding: stderrEncoding ?? this.stderrEncoding
);
```

Then you can still do:

`runOptions.withOptions()..environment['xxx']='yyy';`

but also:

`runOptions.withOptions(workingDirectory: 'x/y/z');`

The name `clone` or `copy` might work too.

lib/grinder_sdk.dart
seaneagan: One reason to keep it as deprecated would be to allow releasing this change without a breaking change version bump, but we have some other important breaking changes that need to happen soon anyways, so I'm not sure if we'll need to release before those or not anyway.  @devoncarew?

devoncarew: I would love to deprecate for now instead of remove. The more we can easy people along API changes the better.

lib/src/run.dart
zoechi: Done. I stuck with `clone` because I think it`s closest to what it does. `cloneAndUpdate` is closer but that seems cumbersome.

lib/grinder_sdk.dart
zoechi: I restored the `workingDirectory` parameter. When I tried two days ago the Analyzer didn't hint about the use of deprecated named arguments, this is why it didn't seem to add value to keep them around, but now I get hints from the analyzer and then it's definitely worth keeping them around a while.

changelog.md
```diff
@@ -1,5 +1,10 @@
 # grinder.dart changes
 
+## UNRELEASED
+- Introduce `RunOptions` to support all `Process.run`/`Process.start`
```
devoncarew: Let's move this down to the 0.7.1 section - we haven't released that yet :)

devoncarew:
In `tool/grind.dart`, ~line 13, can you pass the `--ignore-infos` param into tuneup check? That will unbreak the build. Calling a deprecated method (which we need to do now) is causing a warning.

devoncarew:
@zoechi, thanks for the PR! lgtm w/ a changelog cahnge, build script change, and a green build :)

@seaneagan, you did the most reviewing here - once it looks good to you we can land.

changelog.md
seaneagan: Or maybe just re-label the 0.7.1 section as UNRELEASED?  I like the idea of waiting to give a version number to the "next version" until it's ready to be released, since it's hard to know what the next version number will actually be until you're ready to release, since it depends whether you need to include breaking changes, and if you want to do any dev releases.

lib/src/run.dart
```diff
@@ -95,12 +111,14 @@ Future<String> runAsync(String executable,
     broadcastStderr.listen((List<int> data) => stderr.addAll(data));
     stderrLines.listen(logStderr);
 
+    var encoding = runOptions.stdoutEncoding != null
+        ? runOptions.stdoutEncoding : SYSTEM_ENCODING;
     return process.exitCode.then((int code) {
-      var stdoutString = SYSTEM_ENCODING.decode(stdout);
+      var stdoutString = encoding.decode(stdout);
```
seaneagan: I think we also need to pass `stdoutEncoding` and `stderrEncoding` to the `toLineStream` method, and use it there.

seaneagan: I'd also be fine with just removing `stdoutEncoding` and `stderrEncoding` from `RunOptions` altogether for now.  Not sure they'll actually be needed.

zoechi: done

lib/src/run.dart
```diff
@@ -144,3 +165,66 @@ stderr:
...
+
+/// Helper for methods which support the deprecated [workingDirectory] and the
+/// new [runOptions] to create a [RunOptions] instance which contains the values
+/// of the passed [runOptions] and the passed [workingDirectory].
+/// If both [workingDirectory] and [runOptions.workingDirectory] are passed
+/// an AssertionError is thrown. Only one of both may be used at one
+/// time.
+/// This function can probably be removed when the deprecated `workingDirectory`
+/// arguments are finally removed.
+RunOptions mergeRunOptions(String workingDirectory, RunOptions runOptions) {
```
seaneagan: This should be private i.e. `_mergeRunOptions`.

lib/src/run.dart
```diff
@@ -144,3 +165,84 @@ stderr:
...
+
+/// Helper for methods which support the deprecated [workingDirectory] and the
+/// new [runOptions] to create a [RunOptions] instance which contains the values
+/// of the passed [runOptions] and the passed [workingDirectory].
+/// If both [workingDirectory] and [runOptions.workingDirectory] are passed
+/// an AssertionError is thrown. Only one of both may be used at one
+/// time.
+/// This function can probably be removed when the deprecated `workingDirectory`
+/// arguments are finally removed.
+RunOptions mergeWorkingDirectory(String workingDirectory, RunOptions runOptions) {
```
seaneagan: Let's make this private.

lib/src/run.dart
```diff
@@ -144,3 +165,84 @@ stderr:
...
+
+/// Helper for methods which support the deprecated [envVar] and the
+/// new [runOptions] to create a [RunOptions] instance which contains the values
+/// of the passed [runOptions] and the passed [envVar].
+/// If both [envVar] and [runOptions.environment] are passed
+/// an AssertionError is thrown. Only one of both may be used at one
+/// time.
+/// This function can probably be removed when the deprecated `envVar`
+/// arguments are finally removed.
+RunOptions mergeEnvironment(Map<String,String> environment, RunOptions runOptions) {
```
seaneagan: ...and this.

lib/src/run.dart
zoechi: they are used from two different libraries. I had them private first but then changed it to public. From what I've seen this lib isn't exported from the libs in `lib` anyway therefore it's at least private by convention.

seaneagan: It's exported from here:

https://github.com/google/grinder.dart/blob/master/lib/grinder_tools.dart#L20

zoechi: damn, maybe move them somewhere else. have to check, but it's time for bed now...

lib/src/cli.dart
```diff
@@ -21,7 +21,7 @@ List<String> grinderArgs() => _args;
 List<String> _args;
 bool _verifyProjectRoot;
 
-Future handleArgs(List<String> args, {bool verifyProjectRoot}) {
+Future handleArgs(List<String> args, {bool verifyProjectRoot : false}) {
```
zoechi: I added this default value because a test failed in checked mode because `if(verifyProjectRoot)` with `verifyProjectRoot == null`

lib/src/run.dart
zoechi: I didn't remove `stdoutEncoding` and `stderrEncoding`. Didn't see this comment when I added `done`, but I modified `toLineStream()` to use the encoding set in `RunOptions`. I can imagine this being more useful outside the english speaking area.

zoechi:
Can you initialize a travis test run or should I add a dummy commit? It doesn't look this time the error has anything to do with my code (the previous time it was a failed test which was comprehensible even though it succeeded locally).

seaneagan:
@zoechi lgtm! and thanks for adding tests!

zoechi:
@seaneagan for a good night's sleep ;-)

PR #233, opened 2015-05-15
remove the vm memory options

devoncarew:
Deprecate the `vmNewGenHeapMB` and `vmOldGenHeapMB` vm options. `--new_gen_heap_size` is no longer available, and `--old_gen_heap_size` is I think not as needed.

@seaneagan


seaneagan:
lgtm

I guess `--package-root` will be going away too with the new package spec thing.  I think we can just rely on `vmArgs` going forward.

seaneagan:
changelog update?

devoncarew:
> changelog update?

added -

We may want to keep `--package-root` for a while, to help with the transition, and because tools may still choose to create a single resolved package directory (which the package map file would point into).

PR #235, opened 2015-05-16
apply dartformat

zoechi:


devoncarew:
Cool! It fixed some indent issues - missing a space or one too many spaces.

It also made some method decls and chained list comprehensions harder to read :( I guess the future is standardized formatting though. The Java formatter lets you control line breaking by appending ` //` to the end of a line. If the dart formatter respects that as well we could use that to improve some of the formatter method decls (like the ones using annotations on params).

lgtm! I'll experiment with some formatting tweaks.

zoechi:
Great!

Bob mentioned this trick yesterday on an issue in `dart_style` to avoid moving each entry to a new line in a long list.
```
var list = [ 
  'abc', 'def', 'ghi', 'jkl', //
  'mno', 'pqr', 'stu', 'uvw' //
];
```
Even when moving each value onto its own line is not optimal I found that `dartformat` helps finding an error I'm rather prone to
```
var list = [
  'abc', 'def' 'ghi', 'jkl', 'mno', 'pqr', 'stu', 'uvw', 'xyz','abc', 
  'def' 'ghi', 'jkl', 'mno', 'pqr', 'stu', 'uvw', 'xyz'
];
```
is formatted to 
```
var list = [
  'abc',
  'def' 'ghi',
  'jkl',
  'mno',
  'pqr',
  'stu',
  'uvw',
  'xyz',
  'abc',
  'def' 'ghi',
  'jkl',
  'mno',
  'pqr',
  'stu',
  'uvw',
  'xyz'
];
```

PR #236, opened 2015-05-16
re-format some files

devoncarew:
Add some manual directives to the formater to improve method decl formatting.

@zoechi @seaneagan


zoechi:
LGTM.
I gave up teaching dartformat ;-) 
At the beginning I tried to convince Bob to support annotations to be able to tell `dartformat` to not format parts of the code.
`dartformat` formats most of the code perfect and a few things not perfect but still acceptable, and all this without any cost. In overall quite a good deal.

PR #237, opened 2015-05-16
add a `format` task to tool/grind.dart

zoechi:
With the now formatted Dart code it would make sense to add a `format` task to keep the code formatted.
I don't know if you want it added to the default task.

I prefer passing a concrete list of directories to `dartfmt` because for example with `.` `dartfmt` also formats the `build` directory, if there is any, which then also contains a copy of all dependencies which can make it quite a lengthy but pointless task.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,7 +8,7 @@
   use `RunOptions.workingDirectory` instead.
 - Added `PubApp.runAsync`.
 - In `Dart.run`, deprecate the vmNewGenHeapMB and vmOldGenHeapMB options.
-- Support to pass lists of files or directories to `DartFmt.format()`
+- Support for passing lists of files or directories to `DartFmt.format()`
 
 ## 0.7.0
 - Big changes! Task definititions are now primarily annotation based; see the
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@
   use `RunOptions.workingDirectory` instead.
 - Added `PubApp.runAsync`.
 - In `Dart.run`, deprecate the vmNewGenHeapMB and vmOldGenHeapMB options.
+- Support to pass lists of files or directories to `DartFmt.format()`
 
 ## 0.7.0
 - Big changes! Task definititions are now primarily annotation based; see the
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -8,6 +8,7 @@
   use `RunOptions.workingDirectory` instead.
 - Added `PubApp.runAsync`.
 - In `Dart.run`, deprecate the vmNewGenHeapMB and vmOldGenHeapMB options.
+- Support to pass lists of files or directories to `DartFmt.format()`
```
devoncarew: ==> `Support for passing`

lib/grinder_sdk.dart
```diff
@@ -323,24 +323,37 @@ class Analyzer {
...
 /// the `dartfmt` from SDK 1.9 and greater.
 class DartFmt {
-  /// Run the `dartfmt` command with the `--overwrite` option. Format any files
-  /// in place.
+  /// Run the `dartfmt` command with the `--overwrite` option. Format a file, a
+  /// directory or a list of files or directories in place.
   static void format(fileOrPath) {
-    if (fileOrPath is File) fileOrPath = fileOrPath.path;
-    _run('--overwrite', fileOrPath);
+    if (fileOrPath is File) {
```
devoncarew: How about:

```
if (fileOrPath is File) fileOrPath = fileOrPath.path;
if (fileOrPath is! List) fileOrPath = [fileOrPath];
_run('--overwrite', fileOrPath);
```

tool/grind.dart
```diff
@@ -68,3 +68,14 @@ Future testsBuildWeb() {
...
+
+const sourceDirectories = const [
+  'bin',
+  'example',
+  'lib',
+  'test',
+  'tool',
+  'web'
+];
+@Task('Apply dartformat oo all Dart source files')
```
devoncarew: `oo` ==> `to` :)

devoncarew:
Thanks! Some comments in-line.

lib/grinder_sdk.dart
zoechi: Definitely better.

devoncarew:
lgtm!

Will land this as is. From looking at the API for `Analyzer` though, we have `analyze` (singular)` and `analyzeFiles` (plural). https://github.com/google/grinder.dart/blob/master/lib/grinder_sdk.dart#L310

We should change DartFmt to match or relax the param type on `Analyzer.analyze` and deprecate `Analyzer.analyzeFiles`.

PR #238, opened 2015-05-16
analyzer.analyze

devoncarew:
Extend `Analyzer.analyze` to take a list of files or paths as well.

@seaneagan @zoechi

### DIFF
--- a/lib/grinder_sdk.dart
+++ b/lib/grinder_sdk.dart
@@ -299,7 +299,7 @@ class Dart2js {
  * Utility tasks for invoking the analyzer.
  */
 class Analyzer {
-  /// Analyze a file, a directory or a list of files or directories.
+  /// Analyze a [File], a path ([String]), or a list of files or paths.
   static void analyze(fileOrPaths,
       {Directory packageRoot, bool fatalWarnings: false}) {
     List args = [];
--- a/lib/grinder_sdk.dart
+++ b/lib/grinder_sdk.dart
@@ -299,20 +299,24 @@ class Dart2js {
  * Utility tasks for invoking the analyzer.
  */
 class Analyzer {
-  /// Analyze a single [File] or path ([String]).
-  static void analyze(fileOrPath,
+  /// Analyze a file, a directory or a list of files or directories.
+  static void analyze(fileOrPaths,
       {Directory packageRoot, bool fatalWarnings: false}) {
-    analyzeFiles([fileOrPath],
-        packageRoot: packageRoot, fatalWarnings: fatalWarnings);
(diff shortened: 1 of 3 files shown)


lib/grinder_sdk.dart
```diff
@@ -544,3 +544,12 @@ class _PubLocalApp extends PubApp {
         script: script, arguments: arguments, runOptions: runOptions);
   }
 }
+
+List<String> _coerceToPathList(filesOrPaths) {
+  if (filesOrPaths is! List) filesOrPaths = [filesOrPaths];
```
seaneagan: If `filesOrPaths is File` shouldn't this extract the path e.g. `[filesOrPaths.path]`?

tool/grind.dart
```diff
@@ -14,9 +14,10 @@ void analyze() {
 }
 
 @Task()
-void test() {
-  Tests.runCliTests();
-}
+test() => Tests.runCliTests();
+
+@Task('Apply dartfmt to all Dart source files')
+format() => DartFmt.format(['bin', 'example', 'lib', 'test', 'tool', 'web']);
```
seaneagan: Can this just be `['.']`?

lib/grinder_sdk.dart
seaneagan: Oops, I see your still doing that below, sorry.

lib/grinder_sdk.dart
```diff
@@ -299,20 +299,24 @@ class Dart2js {
  * Utility tasks for invoking the analyzer.
  */
 class Analyzer {
-  /// Analyze a single [File] or path ([String]).
-  static void analyze(fileOrPath,
+  /// Analyze a file, a directory or a list of files or directories.
```
seaneagan: I didn't think directories were supported yet:

https://github.com/dart-lang/analyzer_cli/issues/7

tool/grind.dart
seaneagan: or rather just `'.'`

zoechi: Are you relating to `['bin', 'example', 'lib'...]`. This is intentional because with `'.'` also the `build` directory with all it's subdirectories (contains also copies of all deps) is formatted which is cumbersome and usually not desired.

seaneagan: I see, thanks.  Filed #239 to look at making this easier.

lib/grinder_sdk.dart
devoncarew: Changed to:

> Analyze a [File], a path ([String]), or a list of files or paths.

PR #240, opened 2015-05-18
Devoncarew test 0 12 0

### LINKED ISSUE #58: Integrate with the `test` test runner
Integrate the Dart Test Runner (https://github.com/google/test_runner.dart) in some way.

### LINKED ISSUE #217: upgrade to unit test 0.12.0

devoncarew:
fix #217 
fix #58 

- add a wrapper for the `test` package, called `Test`
- deprecate the `Tests` class
- import both `test` and `unittest` so we can deprecate functionality before removing it
- switch our unit tests over to using `test` and dogfood `Test` in our build script

@seaneagan

### DIFF
--- a/lib/grinder_tools.dart
+++ b/lib/grinder_tools.dart
@@ -61,6 +61,7 @@ void defaultClean([GrinderContext context]) => delete(buildDir);
 /**
  * A utility class to run tests for your project.
  */
+@Deprecated('see `Test`')
 class Tests {
   /**
    * Run command-line tests. You can specify the base directory (`test`), and
@@ -172,6 +173,58 @@ class Tests {
 
       return completer.future;
     });
+  }
+}
+
+/// A wrapper around the `test` package. This class is used to run your unit
+/// tests.
+class Test {
+  final PubApp _test = new PubApp.local('test');
+
+  Test();
+
+  /// Run the tests in the current package. See
+  /// https://pub.dartlang.org/packages/test.
+  ///
+  /// [name] is substring of the name of the test to run. Regular expression
+  /// syntax is supported. [plainName] is a plain-text substring of the name of
+  /// the test to run. [platform] is the platform(s) on which to run the tests.
+  /// Available values are `vm` (default), `dartium`, `content-shell`, `chrome`,
+  /// `phantomjs`, `firefox`, `safari`. [concurrency] controls the number of
+  /// concurrent test suites run (defaults to 4). [pubServe] is the port of a
+  /// pub serve instance serving `test/`.
+  void test({String name, String plainName, List<String> platforms,
+      int concurrency, int pubServe, RunOptions runOptions}) {
+    List<String>args = ['--reporter=expanded'];
+    if (name != null) args.add('--name=${name}');
+    if (plainName != null) args.add('--plain-name=${plainName}');
+    if (platforms != null) args.add('--platform=${platforms.join(',')}');
+    if (concurrency != null) args.add('--concurrency=${concurrency}');
+    if (pubServe != null) args.add('--pub-serve=${pubServe}');
+    _test.run(args, script: 'test', runOptions: runOptions);
+  }
+
+  /// Run the tests in the current package. See
+  /// https://pub.dartlang.org/packages/test.
+  ///
+  /// [name] is substring of the name of the test to run. Regular expression
+  /// syntax is supported. [plainName] is a plain-text substring of the name of
+  /// the test to run. [platform] is the platform(s) on which to run the tests.
+  /// Available values are `vm` (default), `dartium`, `content-shell`, `chrome`,
+  /// `phantomjs`, `firefox`, `safari`. [concurrency] controls the number of
+  /// concurrent test suites run (defaults to 4). [pubServe] is the port of a
+  /// pub serve instance serving `test/`. [color] controls whether the test
+  /// output uses ANSI colors.
+  Future testAsync({String name, String plainName, List<String> platforms,
+      int concurrency, int pubServe, bool color, RunOptions runOptions}) {
+    List<String>args = ['--reporter=expanded'];
+    if (name != null) args.add('--name=${name}');
+    if (plainName != null) args.add('--plain-name=${plainName}');
+    if (platforms != null) args.add('--platform=${platforms.join(',')}');
+    if (concurrency != null) args.add('--concurrency=${concurrency}');
+    if (pubServe != null) args.add('--pub-serve=${pubServe}');
+    if (color != null) args.add(color ? '--color' : '--no-color');
+    return _test.runAsync(args, script: 'test', runOptions: runOptions);
   }
 }
 
--- a/lib/grinder_tools.dart
+++ b/lib/grinder_tools.dart
@@ -197,10 +197,12 @@ class TestRunner {
   /// serving `test/`.
   void test({String name, String plainName, dynamic platformSelector,
       int concurrency, int pubServe, RunOptions runOptions}) {
-    _test.run(
-        _buildArgs(name: name, plainName: plainName, platformSelector: platformSelector,
-                   concurrency: concurrency, pubServe: pubServe),
-        script: 'test', runOptions: runOptions);
+    _test.run(_buildArgs(
+        name: name,
+        plainName: plainName,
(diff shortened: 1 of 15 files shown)


zoechi:
I was thinking about migrating the tests to the new `test` package but you would loose coverage. Coverage doesn't work with `test` yet (AFAIK).

lib/grinder_tools.dart
```diff
@@ -175,6 +176,58 @@ class Tests {
...
 
+/// A wrapper around the `test` package. This class is used to run your unit
+/// tests.
+class Test {
+  final PubApp _test = new PubApp.local('test');
+
+  Test();
+
+  /// Run the tests in the current package. See
+  /// https://pub.dartlang.org/packages/test.
```
seaneagan: will dartdoc make this link clickable?  Maybe use a markdown link:

the [test package](https://pub.dartlang.org/packages/test).

lib/grinder_tools.dart
```diff
@@ -175,6 +176,58 @@ class Tests {
...
+  /// [name] is substring of the name of the test to run. Regular expression
+  /// syntax is supported. [plainName] is a plain-text substring of the name of
+  /// the test to run. [platform] is the platform(s) on which to run the tests.
+  /// Available values are `vm` (default), `dartium`, `content-shell`, `chrome`,
+  /// `phantomjs`, `firefox`, `safari`. [concurrency] controls the number of
+  /// concurrent test suites run (defaults to 4). [pubServe] is the port of a
+  /// pub serve instance serving `test/`.
+  void test({String name, String plainName, List<String> platforms,
+      int concurrency, int pubServe, RunOptions runOptions}) {
+    List<String> args = ['--reporter=expanded'];
```
seaneagan: Move this and the next five lines to a utility function that is used here and in `testAsync`?

lib/grinder_tools.dart
```diff
@@ -175,6 +176,58 @@ class Tests {
...
+  /// output uses ANSI colors.
+  Future testAsync({String name, String plainName, List<String> platforms,
+      int concurrency, int pubServe, bool color, RunOptions runOptions}) {
+    List<String> args = ['--reporter=expanded'];
+    if (name != null) args.add('--name=${name}');
+    if (plainName != null) args.add('--plain-name=${plainName}');
+    if (platforms != null) args.add('--platform=${platforms.join(',')}');
+    if (concurrency != null) args.add('--concurrency=${concurrency}');
+    if (pubServe != null) args.add('--pub-serve=${pubServe}');
+    if (color != null) args.add(color ? '--color' : '--no-color');
```
seaneagan: Why is `color` only supported for `testAsync` and not `test`?

lib/grinder_tools.dart
```diff
@@ -175,6 +176,58 @@ class Tests {
...
+    _test.run(args, script: 'test', runOptions: runOptions);
+  }
+
+  /// Run the tests in the current package. See
+  /// https://pub.dartlang.org/packages/test.
+  ///
+  /// [name] is substring of the name of the test to run. Regular expression
+  /// syntax is supported. [plainName] is a plain-text substring of the name of
+  /// the test to run. [platform] is the platform(s) on which to run the tests.
+  /// Available values are `vm` (default), `dartium`, `content-shell`, `chrome`,
```
seaneagan: Can we just link to the test packages documentation on this instead of inlining it here:

https://github.com/dart-lang/test#platform-selector-syntax

It would also be nice to link to test package for the other docs as well, but those don't seem to have documentation in the README.

lib/grinder_tools.dart
```diff
@@ -175,6 +176,58 @@ class Tests {
...
+  /// `phantomjs`, `firefox`, `safari`. [concurrency] controls the number of
+  /// concurrent test suites run (defaults to 4). [pubServe] is the port of a
```
seaneagan: Can we just link here:

https://github.com/dart-lang/test#testing-with-barback

lib/grinder_tools.dart
```diff
@@ -175,6 +176,58 @@ class Tests {
...
+  /// pub serve instance serving `test/`.
+  void test({String name, String plainName, List<String> platforms,
```
seaneagan: I think we should support `platforms` being a `String` either instead of or in addition to `List<String>` to get the full power of the [platform selector syntax](https://github.com/dart-lang/test#platform-selector-syntax).  I think once https://github.com/dart-lang/test/issues/23 is fixed, it will probably expose an enum instead, so that'd be one reason to maybe leave out the `List<String>` option for now until that shakes out.

seaneagan: also maybe `platformSelector` may be more self-documenting in at least in the case of `String`.

lib/grinder_tools.dart
```diff
@@ -175,6 +176,58 @@ class Tests {
   }
 }
 
+/// A wrapper around the `test` package. This class is used to run your unit
+/// tests.
+class Test {
```
seaneagan: I like that `Test` matches the name of the package (`test`), but without reading the documentation for it, I'd assume `Test` represents an individual unit test, or run thereof, not a test running API.  Also once https://github.com/dart-lang/test/issues/23 is fixed, that will expose just such a `Test` class, and that may become the preferred way to run tests from grinder since it avoid shelling out, but we would likely have a name conflict.  Some other suggestions would be `Tester`, `TestRunner`, or `TestApp`.  @zoechi thoughts?

lib/grinder_tools.dart
```diff
@@ -61,6 +61,7 @@ void defaultClean([GrinderContext context]) => delete(buildDir);
 /**
  * A utility class to run tests for your project.
  */
+@Deprecated('see `Test`')
```
seaneagan: I thnk we should start using e.g. `'see [Test]'` instead pending the outcome of:

https://github.com/dart-lang/dartdoc/issues/500

but that can be a separate PR to update all of our deprecation messages.

lib/grinder_tools.dart
seaneagan: Also, we could look at adding a --color option to grinder itself, and default this one to that, or maybe even remove this one?

seaneagan: See #243.

lib/grinder_tools.dart
zoechi: I think `TestRunner` is fine.

lib/grinder_tools.dart
devoncarew: done

lib/grinder_tools.dart
devoncarew: sgtm

lib/grinder_tools.dart
devoncarew: Ah, an oversight. I like your idea of keying off a global color property. I'll remove this param.

lib/grinder_tools.dart
devoncarew: I added the link, but want to keep the inlined info, which I think will help users use the class.

lib/grinder_tools.dart
devoncarew: I think this will be fairly infrequently used, I just want to expose it for the sake of completeness.

lib/grinder_tools.dart
devoncarew: Cool, changed to `dynamic platformSelector`; it can take a list or a string.

lib/grinder_tools.dart
devoncarew: kk, `TestRunner` sounds good; I'll document that it maps to the `test` package.

zoechi: but `Test` is fine for me too. No strong opinion here.

lib/grinder_tools.dart
devoncarew: Makes sense; changed to `@Deprecated('see [TestRunner]')`.

devoncarew:
Updated - thanks for the comments!

seaneagan:
lgtm!

Can we also now deprecate/remove the browser driving code in grinder?

devoncarew:
> Can we also now deprecate/remove the browser driving code in grinder?

sounds good; let's deprecate until 0.8.0

PR #246, opened 2015-05-19
add a files param to TestRunner.test

devoncarew:
- add a `files` param to `TestRunner.test`
- publish a patch version

### DIFF
--- a/lib/grinder_sdk.dart
+++ b/lib/grinder_sdk.dart
@@ -13,6 +13,7 @@ import 'package:which/which.dart';
 import 'grinder.dart';
 import 'src/run.dart' as run_lib;
 import 'src/run_utils.dart';
+import 'src/utils.dart';
 
 bool _sdkOnPath;
 
@@ -305,7 +306,7 @@ class Analyzer {
     List args = [];
     if (packageRoot != null) args.add('--package-root=${packageRoot.path}');
     if (fatalWarnings) args.add('--fatal-warnings');
-    args.addAll(_coerceToPathList(fileOrPaths));
+    args.addAll(coerceToPathList(fileOrPaths));
     run_lib.run(_sdkBin('dartanalyzer'), arguments: args);
   }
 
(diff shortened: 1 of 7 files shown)


lib/grinder_tools.dart
```diff
@@ -217,9 +225,11 @@ class TestRunner {
   /// `firefox`, `safari`. [concurrency] controls the number of concurrent test
   /// suites run (defaults to 4). [pubServe] is the port of a pub serve instance
   /// serving `test/`.
-  Future testAsync({String name, String plainName, dynamic platformSelector,
-      int concurrency, int pubServe, RunOptions runOptions}) {
+  Future testAsync({dynamic files, String name, String plainName,
+      dynamic platformSelector, int concurrency, int pubServe,
+      RunOptions runOptions}) {
     return _test.runAsync(_buildArgs(
+        files: files,
```
seaneagan: An optional positional parameter would be more consistent with the named command-line test runner and with `Dartfmt` and `Analyzer`, but we'd have to wait until [this DEP](https://github.com/lrhn/dep-optionals) until we could actually make it optional.  If we keep it named, maybe `paths` is more self-documenting that you can use a directory path and not just file paths?

devoncarew: Totally agree re: the optional positional parameter; would have preferred to have something like that available. I think the primary use-case will be files however; would prefer to keep it as is.

seaneagan: sgtm

lib/grinder_sdk.dart
```diff
@@ -305,7 +306,7 @@ class Analyzer {
     List args = [];
     if (packageRoot != null) args.add('--package-root=${packageRoot.path}');
     if (fatalWarnings) args.add('--fatal-warnings');
-    args.addAll(_coerceToPathList(fileOrPaths));
+    args.addAll(coerceToPathList(fileOrPaths));
```
zoechi: It's a bit weird that here are `fileOrPaths`, `files` (line 320), `fileOrPath` (line 335) for the same kind of argument. Also the doc-comments are written differently.

PR #255, opened 2015-05-24
Update changelog.md

sethladd:


zoechi:
Changelog isn't shown at pub.dartlang.org. Should this be fixed in pub or here CHANGELOG.md instead of Changelog.md

devoncarew:
We can change grinder's to `CHANGELOG.md` from `changelog.md`. The all caps is more typical. It would be nice if pub.dartlang.org was case insensitive thought -

sethladd:
this might be a regression. They just moved over to the Dart version of pub.dartlang.org. I'll open an issue over there.

sethladd:
https://github.com/dart-lang/pub-dartlang-dart/issues/17

Issue #239, opened 2015-05-18
Expose list of common dart source dirs as targets for formatter, linter, analyzer, etc.


seaneagan:
It'd be nice to default `Analyzer`, `Dartfmt` (others?) to running on all common dart source directories.  If we had [this DEP](https://github.com/lrhn/dep-optionals) then we could just make the paths optional in `Analyzer.analyze` and `Dartfmt.format/dryRun`, but maybe more generally useful would be exposing the list of dirs:

```dart
final List<Directory> sourceDirs = ['bin', 'example', 'lib', 'test', 'tool', 'web'].map((path) => new Directory(path)).toList();
```

or `dartSourceDirs`?

`web` can contain dart source code, right?

zoechi:
Yes, `web` usually contains `dart` source. 
I haven't tried if analyzer or formatter choke on not existing directories.
`sourceDirs.where((d) => d.existsSync())` isn't too hard, maybe adding directly?
Having such a list is handy, I usually reuse it in several tasks.

seaneagan:
I like it!

```
final List<Directory> sourceDirs = ['bin', 'example', 'lib', 'test', 'tool', 'web'].map((path) => new Directory(path)).where((d) => d.existsSync()).toList();
```

devoncarew:
Hmm, maybe a static list of top-level dirs that can contain dart source, and a method that will return the subset that actually exist for the current project? So,

```dart
const List<String> sourceDirs;
List<String> getExistingSourceDirs();
```

seaneagan:
sgtm, though I'd make both getters and use `Directory` instead of `String`:

```dart
const List<Directory> sourceDirs;
List<Directory> get existingSourceDirs => sourceDirs.where((d) => d.existsSync()).toList();
```

seaneagan:
or probably more performant:

```dart
List<Directory> get existingSourceDirs => Directory.current.listSync().where(sourceDirs.contains).toList();
```

devoncarew:
I generally don't think of getters as things that could throw exceptions.

seaneagan:
The style guide doesn't mention that as a criteria:

https://www.dartlang.org/articles/style-guide/#do-use-a-getter-for-operations-that-conceptually-access-a-property

is it a Java thing?  I don't personally see a need for that distinction.

zoechi:
Fixed by #260

PR #260, opened 2015-05-28
- Auto-expand directories passed to `Analyzer.analyze()` to a list of Dart source files

zoechi:


zoechi:
The analyzer tests are quite slow on my system, maybe I should just bunch them together to from 10 to 2, but usually finer grained tests are better because when failing they immediately show what went wrong.

lib/grinder_sdk.dart
```diff
@@ -17,6 +18,17 @@ import 'src/utils.dart';
 
 bool _sdkOnPath;
 
+// Returns the common top-level directories which usually contain Dart source
+// and which actually exist in the current project.
+Set<String> get defaultSourceDirectories => [
```
seaneagan: I don't think `default` is necessary.  And we already have e.g. `binDir`, so we should similarly abbreviate `Directories` here.  Also, I think we wanted to have separate fields for existing vs. all, see discussion in #239. So I'd suggest:

```
Set<Directory> sourceDirs = ['bin', 'example', 'lib', 'test', 'tool', 'web'].map((path) => new Directory(path)).toSet();
Set<Directory> get existingSourceDirs => Directory.current.listSync().where(sourceDirs.contains).toSet();
```

test/grinder_sdk_test.dart
```diff
@@ -128,4 +128,46 @@ main() {
       expect(newText, isNot(equals(originalText)));
     });
   });
+
+  group('grinder.sdk Analyzer', () {
```
seaneagan: I think many of these tests should instead be tests of `coerceToPathList` [here](https://github.com/google/grinder.dart/blob/master/test/src/utils_test.dart).

seaneagan: Or of `_expandDirectoriesToDartFiles`.

lib/grinder_sdk.dart
```diff
@@ -339,6 +352,41 @@ class Analyzer {
...
+  }
+
+  static List<String> _collectDartFiles(Directory directory) {
+    var files = <String>[];
+    if (directory.existsSync()) {
+      for (var entry
+          in directory.listSync(recursive: true, followLinks: false)) {
+        //var relative = path.relative(entry.path, from: directory.path);
+
+        if (_isDartFileName(entry.path) && !_isInHiddenDir(directory.path)) {
```
seaneagan: `directory.path` -> `entry.path`?

lib/grinder_sdk.dart
```diff
@@ -339,6 +352,41 @@ class Analyzer {
...
+          files.add(entry.path);
+        }
+      }
+    }
+    return files;
+  }
+
+  /// Returns [:true:] if this [fileName] is a Dart file.
```
seaneagan: [:true:]

is the legacy format, and is not standard markdown.  I prefer backticks:

    `true`

lib/grinder_sdk.dart
zoechi: Thanks for pointing that out. I thought it was just a comment on another issue. 
`Directory` seems not to have `==`, `identical` seems a bit overkill, and I don't like that one returns realtive dirs, while the other returns absolute ones.
This looks a bit overcomplicated but it was the simplest I found
```
Set<Directory> get existingSourceDirs => Directory.current
    .listSync()
    .where((f) => f is Directory)
    .map((d) => new Directory(path.relative(d.path)))
    .where((d) => sourceDirs.any((sd) => sd.path == d.path))
    .toSet();
```

lib/grinder_sdk.dart
zoechi: Don't remember this been changed. Thanks for the info.

lib/grinder_sdk.dart
zoechi: :+1:

pubspec.yaml
```diff
@@ -17,6 +17,7 @@ dependencies:
   cli_util: '>=0.0.1 <0.1.0'
   glob: '>=1.0.0 <2.0.0'
   http_server: '>=0.9.0 <0.10.0'
+  path: '^1.3.5'
```
devoncarew: Can we make this version constrain broader? `path` is heavily used, and I want to make sure we keep our constraints wide so that we don't contribute to unresolvable dependencies.

Perhaps `^1.0`?

lib/grinder_sdk.dart
```diff
@@ -17,6 +18,21 @@ import 'src/utils.dart';
 
 bool _sdkOnPath;
 
+Set<Directory> sourceDirs = [
```
devoncarew: `const` or `final`?

devoncarew: And perhaps some (short) dartdoc for this fields and the getter below.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # grinder.dart changes
 
+## UNRELEASED
+- Auto-expand directories passed to `Analyzer.analyze()` to a list of Dart 
+  source files using the added method `findDartSourceFiles()`.
+- Modified `coerceToPathList()` to accept `Iterable` instead of `List`
```
devoncarew: We probably don't need to mention `coerceToPathList` in the release notes since it's not part of the public api.

pubspec.yaml
zoechi: Good point

devoncarew:
lgtm!

PR #268, opened 2015-06-09
- #265 return value instead of InstanceMirror

### LINKED ISSUE #265: Weird issue with `InstanceMirror of _Future` returned from `task.execute()`
Some explanation why I do what I do:
I have a package with Grinder tasks I share between different packages.
To be able to override specific tasks I use the following strategy

```Dart
@Task('Do something')
some() => someTask();

Function someTask = someTaskImpl;

someTaskImpl() => doSomethingAsync();
```
This looks a bit cumbersome but it's the easiest way I found to be able to customize single tasks while still being able to use the tasks in `@Depends(  ...needsToBeConst... )` annotations.

A default `tool/grind.dart` has just this line

```Dart
export 'package:bwu_utils_dev/grinder/default_tasks.dart' hide main;
```
or if I want to customize a few more
```Dart
library bwu_utils_dev.tool.grind;

export 'package:bwu_utils_dev/grinder/default_tasks.dart' hide main;
import 'package:bwu_utils_dev/grinder/default_tasks.dart' hide main;

main(List<String> args) {
  someTask = () => print('Hello, World');
  grind(args);
}
```

Now I run into the problem that `var result = task.execute(context);` returns `InstanceMirror of _Future` which results in Grinder not waiting for the async execution to complete.

It doesn't look like this is caused by Grinder (I might have overlooked something of course). To me it looked more like a zones bug but wasn't able to track the root cause down.
Changing `_executeTask` like below fixed it for me. 
I'll create a pull request but maybe you might have a better solution.

(code from https://github.com/google/grinder.dart/blob/master/lib/grinder.dart#L400)

```Dart
Future _executeTask(GrinderTask task) {
    log('${task}');

    GrinderContext context = new GrinderContext._(this, task);
    var result = task.execute(context);

    if (!(result is Future)) {
      if (result is InstanceMirror && result.reflectee is Future) {
        result = result.reflectee;
      } else {
        result = new Future.value(result);
      }
    }

    // TODO: whenComplete(), dispose of the context?
    return (result as Future).then((_) {
      log('');
    });
  }
```

zoechi:
- fix #265
- add tests

### DIFF
--- a/test/src/discover_tasks_test.dart
+++ b/test/src/discover_tasks_test.dart
@@ -12,6 +12,7 @@ import 'package:test/test.dart';
 import 'task_discovery/good_tasks.dart' as good;
 import 'task_discovery/bad_tasks.dart' as bad;
 import 'task_discovery/external_tasks.dart' as external;
+import '_common.dart';
 
 main() {
   // Libs which contains annotated tasks (imported above).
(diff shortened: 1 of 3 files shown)


test/src/discover_tasks_test.dart
```diff
@@ -147,6 +148,40 @@ main() {
     });
   });
 
+  group('execute returns value', () {
+    grinderTest('task from method', () {}, (MockGrinderContext ctx) {
+      var annotated =
+          discoveryGood.discoverDeclaration(goodLib.declarations[#method], {});
+      var task = annotated.task;
+      expect(task.name, 'method');
```
seaneagan: There are already tests for the names of discovered tasks, so can we remove those here?

zoechi:
The AppVeyor issue looks similar to https://github.com/dart-lang/test/commit/57355ed7e6067b00d401e5feabfbe95f6b431617

test/src/discover_tasks_test.dart
zoechi: The intention was not to test the discovery but whether the return value was handled properly. I tried to cover the three possible paths.
At first I intended to just add two lines (execute the discovered tasks and evaluate the result) where discovery was tested, but with the need of `GrinderContext` this seemed to add too much noise and disguise the main intent of the tests. 
Is there a way to simplify these added tests? I haven't played with imperatively adding tasks to a `Grinder` instance yet.

seaneagan: As far as I can tell, the tests look good.  They are testing the the correct `taskFunction` is generated based on the declaration.  Except no need to test the names, since that's already tested elsewhere.

We could also add tests that the return value from arbitrary `taskFunction`s are used, and to test that `Grinder.start` properly waits for futures returned from `GrinderTask.execute`.  But no need to do that in this PR.

devoncarew:
lgtm!

Hopefully the build failure (https://ci.appveyor.com/project/devoncarew/grinder-dart/build/1.0.317) will be addressed by the next version of `test`.

test/src/discover_tasks_test.dart
zoechi: [This test](https://github.com/google/grinder.dart/blob/master/test/grinder_test.dart#L197) already checks that `grinder.start` waits for a future.

PR #269, opened 2015-06-11
Devoncarew remove tests impl

devoncarew:
- prefer `test` to run web (and cli) tests
- remove the dep on `webkit_inspection_protocol`
- have `Tests. runWebTests` fail with a message pointing to `TestRunner`
- remove unused and unreferenced code


devoncarew:
AppVeyor build is unhappy; hopefully fixed w/ an new release of `test.

@seaneagan for review

seaneagan:
lgtm!  Removal of the web testing stuff is a breaking change, but are we assuming nobody is using it at this point?

devoncarew:
Yeah, that's what I'm assuming. It's a breaking change to the functionality, but it'll only impact a very small percentage of users (possibly zero!). And not reving the major version is a nice benefit to downstream users.

Issue #278, opened 2015-06-24
create a new release


devoncarew:
We have some fixes we need to capture...

devoncarew:
Possibly fix https://github.com/google/grinder.dart/issues/273 first?

devoncarew:
pushed v0.7.2

Issue #138, opened 2015-03-27
Remove things that were deprecated in 0.7.0


seaneagan:
Lots of things are deprecated in 0.7.0.  We should remove these in the next breaking release.

devoncarew:
We may want to remove the `GrinderContext context` params in the tools APIs; I think this'll make the APIs nicer. It _will_ be a breaking change, but should be easy for users to adjust to.

devoncarew:
largely done

PR #281, opened 2015-06-30
Implement task parameterization

seaneagan:
@devoncarew 

This allows defining parameters for tasks, and accepting arguments for them when invoking tasks either from the command-line or when used as dependencies.

It doesn't yet include:

* An annotation based API (want to factor out that code from unscripted so that it can be shared here)
* README and CHANGELOG updates (waiting for the annotation based API first)
* Help output for task parameters
* Tab-completion of positional and named values (option names should already be completed though)
* Re-exporting unscripted's `Option`, `Flag`, `Positional`, and `Rest` (hesitant to do this).
* Optional positionals are only supported via the `Rest` argument (multiple homogeneous args).  Main reason is the same as it is for unscripted, which is that dart doesn't support having both named and positional optional parameters in the same signature, so they couldn't be defined through the forthcoming annotation based API.  There may be a way to add it in the imperative API though.

I will file separate issues for each of these, once this one lands.

This PR also splits out most of `lib/grinder.dart` into smaller libraries under `lib/src`.  I didn't intend on doing that in this PR, but I added a couple new libraries, and wanted to use just one small part of `lib/grinder.dart` so factored that part out, but then that depended on another small part, so had to factor that out as well, and it snowballed from there.


lib/src/cli_util.dart
```diff
@@ -0,0 +1,267 @@
+// Copyright 2015 Google. All rights reserved. Use of this source code is
```
seaneagan: Some of the methods in this file are not yet directly unit tested.  That's because they are mostly copied and pasted from unscripted with slight modifications, and so I want to factor them out into a separate library or package that can be shared by unscripted and grinder.  Plan to do that as part of implementing the annotation based API.

lib/src/grinder.dart
```diff
@@ -0,0 +1,224 @@
...
+        Duration elapsed = new DateTime.now().difference(startTime);
+        log('finished in ${elapsed.inMilliseconds / 1000.0} seconds.');
+      });
+    } else {
+      return new Future.value();
+    }
+  }
+
+  /// Given a task, return all of its immediate dependencies.
+  Iterable<TaskInvocation> getImmediateDependencies(GrinderTask task) =>
```
seaneagan: This is a breaking change, used to return `Iterable<GrinderTask>`.  But I doubt it is used by any grinder consumers.  Yet, I think a minor version bump would be the right thing to do.

lib/src/grinder.dart
```diff
@@ -0,0 +1,224 @@
...
+  task.depends;
+
+  /// Given a task, return all of its transitive dependencies.
+  List<TaskInvocation> getAllDependencies(GrinderTask task) => _taskDeps[task];
```
seaneagan: Also a breaking change, used to return `List<GrinderTask>`.

tool/grind.dart
```diff
@@ -5,8 +5,28 @@ import 'dart:io';
 
 import 'package:grinder/grinder.dart';
 
-main(args) => grind(args);
+import 'package:unscripted/unscripted.dart';
 
+main(args) {
+
+  addTask(new GrinderTask(
```
seaneagan: I basically just added this task as an example of how to use the imperative API.  Though this task could be useful if we also make it update the version constant in the code simultaneously.  This is also the use case @sethladd was wanting this feature for.

devoncarew:
It is a doozy - sorry for the delay in taking a look at it.

> This PR also splits out most of lib/grinder.dart into smaller libraries under lib/src

This does make the PR harder to review - it's not easy to tell at a glance which changes are semantic ones...

That aside, given the size of the PR, the source code re-org, and the fact that it has some smaller breaking changes, I think we should push out a last minor release before this lands, and then plan on doing a full major rev. It'll capture this PR, give you some time to do the additional changes mentioned, and let us remove the 0.7 deprecated code.

lib/src/grinder.dart
devoncarew: Hmm, how does a task differ from a `TaskInvocation`? Currently grinder builds up set of dependencies, so we can know that `a` depends on `b`. Would we still know that, or can `a` now only depend on an invocation of `b`?

It would be nice to still know that logically, `a` depends on `b`, but that for this run, the invocations `a` and `b(foo)` have been requested.

tool/grind.dart
```diff
@@ -5,8 +5,28 @@ import 'dart:io';
...
+      'bump',
+      taskFunction: () {
+        var releaseType = context.invocation.positionals.first;
```
devoncarew: Can we make accessing the args more terse, or add aliases to help references to the args be terse?

```
context.args.rest.first;
context.args.first;
context.args['foo];

args.rest.first;
args['foo'];
```

thoughts? This: `context.invocation.positionals.first` strikes me as pretty long, given that it may be the most common use of task args.

lib/src/grinder.dart
seaneagan: A `TaskInvocation` is an invocation of a task, so it contains the name of the task attempting to be executed, and which args (positionals and named) to pass to the task of that name.

So if task `a` depends on task `b` and doesn't need to pass it any args (the most common case), then this reflects as a `a` having a dependency on `const TaskInvocation('b')`, which is the same as `const TaskInvocation('b', positionals: const [], options: const {})`.  But it's now also possible for task `a` to depend on e.g. `const TaskInvocation('b', positionals: const ['some positional'], options: const {'some-option': 'some value'})`.

However, it's only allowed to invoke a given task once (with a single set of args), in any grinder run.  And there's a test for that.

Does that make sense?

tool/grind.dart
seaneagan: I don't anticipate many folks actually needing to use the imperative API introduced in this PR, once I add the follow up annotation-based API which will look something like:

```dart
@Task('bump the version')
bump(
    @Positional(help: 'Which type of release to do', allowed: ['release', 'build', 'patch', 'minor', 'major'])
    String releaseType,
    {bool pre: false
     String preId}) {
  // ...
}
```

which is about as terse as it could possibly get.

The name `invocation` is to match `TaskInvocation`.  I considered `TaskArgs`, which would match up with `args`, but that doesn't really capture the fact that it contains the task name as well.  And "invocation" is already a familiar concept for anyone who has used `dart:mirrors`.

seaneagan:
> This does make the PR harder to review - it's not easy to tell at a glance which changes are semantic ones...

Yep, totally my fault.  `lib/grinder.dart` was getting pretty unwieldy, so I added the new libraries separately, and then my habit of only depending on the parts I need from those libraries kicked in, and I didn't realize I was breaking the PR diffs until it was too late.  FWIW, it was really nice working with the smaller libraries compared to endless scrolling up and down `lib/grinder.dart` trying to find things :).

> That aside, given the size of the PR, the source code re-org, and the fact that it has some smaller breaking changes, I think we should push out a last minor release before this lands, and then plan on doing a full major rev. It'll capture this PR, give you some time to do the additional changes mentioned, and let us remove the 0.7 deprecated code.

Sounds good, so first #278 then merge this, then #138, then the follow up issues I mentioned above, then release 0.8?

lib/src/grinder.dart
devoncarew: Cool, thanks for the explanation. I think my concern here was that we're mixing tasks and task invocations. So we're querying if Task a depends on TaskInvocation b, instead of Task a depending on Task b or TaskInvocation a depending on TaskInvocation b.

tool/grind.dart
devoncarew: Ah, ok, cool. The 95% case will be the annotations API, which will be nice and terse; makes sense.

devoncarew:
> Sounds good, so first #278 then merge this, then #138, then the follow up issues I mentioned above, then release 0.8?

+1!

devoncarew:
OK, lgtm!

PR #311, opened 2016-10-14
Use Platform.resolvedExecutable.

nex3:
This ensures that everything runs using the same Dart version as
everything else, and that the user can control this by choosing which
version of Pub they run.


devoncarew:
lgtm, thanks!

PR #348, opened 2020-03-18
analysis_options v1.9.0 compliance

domesticmouse:
PTAL @devoncarew

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # grinder.dart changes
 
+## 0.8.4
+- Update the code to be analysis_options v1.9.0 compliant
+
 ## 0.8.3+1
 - Mark `fail()` and `GrinderContext.fail()` as `@alwaysThrows`.
 - Use `package:pedantic` to analyze the package.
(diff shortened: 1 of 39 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # grinder.dart changes
 
+## 0.8.4
+- Update the code to be analysis_options v1.9.0 compliant
```
devoncarew: ==> `package:pedantic v1.9.0 compliant`

example/grind.dart
```diff
@@ -3,21 +3,21 @@
 
 import 'package:grinder/grinder.dart';
 
-main(args) => grind(args);
+Future<dynamic> main(args) => grind(args);
 
 @Task('Initialize stuff.')
-init() {
-  log("Initializing stuff...");
+void init() {
```
devoncarew: I like the added use of types on API boundaries; definitely miss the use of types within a method.

CHANGELOG.md
domesticmouse: Much better, thanks Devon. Done.

example/grind.dart
domesticmouse: Yeah, agree

PR #350, opened 2020-06-09
0.8.5: broaden cli_util dep

pq:
/cc @devoncarew 

/fyi @srawlins


pq:
Thanks!  Can you publish too?  (Or add me to the publisher's list?)

devoncarew:
I just published, but will also add you as a publisher.

PR #357, opened 2020-10-12
Declare a Never return type for fail() and GrinderContext.fail()

nex3:


pubspec.yaml
```diff
@@ -1,12 +1,12 @@
...
 # This version must be updated in tandem with the lib/src/cli.dart version.
-version: 0.8.5
+version: 0.8.6
 description: Grinder is a task runner for Dart, helping to define and automate common project workflows.
 
 homepage: https://github.com/google/grinder.dart
 
 environment:
-  sdk: '>=2.3.0 <3.0.0'
+  sdk: '>=2.8.1 <3.0.0'
```
devoncarew, line 9: It looks like this passed on the CI - the use of `Never`. The CI is running `2.11.0-190.0.dev`; I'm trying to confirm when Never is available, but it might not be until 2.11.

We should test this against the current stable (`2.10`) to see if it passes; if so, we could increase the lower sdk bounds here.

Other options would be to rev the sdk lower bounds to 2.11-dev, or, use `Null` here in place of Never.

nex3: I tested this locally with 2.10 and it worked great.

devoncarew: Can you bump the constraint here? I'll update the CI to test against both dev and stabvle.

devoncarew: https://github.com/google/grinder.dart/pull/358

nex3: Done.

devoncarew (approved):

devoncarew:
I may have to revert this PR; I'm seeing these issues locally:

```
Analyzing grinder.dart...              3.4s

   info • The type Never isn't yet supported at lib/src/grinder_context.dart:40:3 • (sdk_version_never)
   info • The type Never isn't yet supported at lib/src/grinder_context.dart:59:1 • (sdk_version_never)
   info • The type Never isn't yet supported at lib/src/singleton.dart:26:3 • (sdk_version_never)
   info • The type Never isn't yet supported at test/src/_common.dart:40:3 • (sdk_version_never)

4 issues found.
```

PR #361, opened 2020-12-08
Support running in null-safety mode

nex3:


devoncarew:
Thanks! I'll review this over the course of the day.

lib/grinder_files.dart
```diff
@@ -336,13 +339,12 @@ Directory getDir(String path) {
 }
 
 void copy(FileSystemEntity entity, Directory destDir,
-    [GrinderContext context]) {
+    [GrinderContext? context]) {
```
devoncarew, line 342: It looks like the `context` arg was never used in `_copyImpl()`. We should mark this param as deprecated.

lib/src/cli.dart
```diff
@@ -15,16 +15,19 @@ import 'singleton.dart' as singleton;
 // This version must be updated in tandem with the pubspec version.
 const String appVersion = '0.8.5';
 
-List<String> grinderArgs() => _args;
-List<String> _args;
-bool _verifyProjectRoot;
+List<String> grinderArgs() {
+  var args = _args;
+  if (args == null) fail('grinderArgs() may only be called after grind().');
```
devoncarew, line 20: Instead of assigning to a local we could also check `_args` and after that return `_args!`. That would be slightly more terse.

lib/src/cli.dart
```diff
@@ -15,16 +15,19 @@ import 'singleton.dart' as singleton;
...
+
+List<String>? _args;
 
 Future runTasks(
   List<String> args, {
   bool verifyProjectRoot = false,
 }) async {
-  _args = args ?? [];
-  _verifyProjectRoot = verifyProjectRoot;
+  _args = args;
```
devoncarew, line 30: I don't know about actual uses in the wild, but it looks like this method was handling cases where the `args` param could be null.

lib/src/discover_tasks.dart
```diff
@@ -5,6 +5,8 @@ library grinder.src.discover_tasks;
 
 import 'dart:mirrors';
 
+import 'package:collection/collection.dart';
```
devoncarew, line 8: Note that we don't have a direct dep on `collection` in our pubspec file.

lib/src/grinder.dart
```diff
@@ -19,16 +20,16 @@ import 'task_invocation.dart';
 void addTask(GrinderTask task) => grinder.addTask(task);
 
 /// The default task run when no tasks are specified on the command line.
-GrinderTask get defaultTask => grinder.defaultTask;
+GrinderTask? get defaultTask => grinder.defaultTask;
 
-set defaultTask(GrinderTask v) {
+set defaultTask(GrinderTask? v) {
```
devoncarew, line 25: Do we want the setter to support nullability?

devoncarew: Ah, from below, it's because this needs to have the same type as the getter.

lib/src/discover_tasks.dart
devoncarew: OK, I see this is added in this PR.

test/src/run_script.dart
```diff
@@ -1,3 +1,4 @@
+// @dart=2.9
```
devoncarew, line 1: nit: I believe this comment can be before the imports as well.

devoncarew (approved):
Thanks for the PR (and sorry for the delay in review)!

I have a few comments in-line. I'll land this PR and address any items in a follow up, once I have a chance to play with this locally.

PR #374, opened 2021-03-17
Fix process exception in windows

liudonghua123:
try to fix https://github.com/dart-lang/sdk/issues/37751#issuecomment-800369852, https://github.com/dart-lang/sdk/issues/39455.

PR #375, opened 2021-03-18
try to fix grind check-init failure

liudonghua123:
see https://github.com/google/grinder.dart/pull/374.


devoncarew (approved):
Thanks! Great diagnoses and fix.

PR #377, opened 2021-03-20
update grinder dependency to the current

liudonghua123:
I send a pr about fixing `check-init` (https://github.com/google/grinder.dart/pull/375) grinder task, however the grinder dependency is not reference the current grinder project. So I make a patch about this.


devoncarew (approved):
👍

PR #380, opened 2021-06-02
Update dependencies and switch to recommended lints

parlough:


lib/grinder_sdk.dart
```diff
@@ -41,7 +41,7 @@ final Directory sdkDir =
 /// This is deprecated.
 ///
 /// Use [sdkDir] instead.
-@deprecated
+@Deprecated('Use sdkDir to get the current SDK directory instead.')
```
devoncarew, line 44: 👍 these are good improvements :)

devoncarew (approved):
Thanks for the contribution!

Issue #385, opened 2022-03-17
Pub broken on Dart 2.17


jathak:
The standalone `pub` tool is being removed in Dart 2.17, which causes any usage of the [`Pub` class](https://pub.dev/documentation/grinder/latest/grinder.sdk/Pub-class.html) to fail.

References to the standalone `pub` tool should be replaced with `dart pub`.

liudonghua123:
I have the similar issue.

```shell
D:\code\dart\app_launcher>dart --version
Dart SDK version: 2.17.0-246.0.dev (dev) (Sun Mar 27 12:20:08 2022 -0700) on "windows_x64"

D:\code\dart\app_launcher>
D:\code\dart\app_launcher>dart tool/grind.dart clean build
grinder running clean test build

clean

test
  D:\apps\dart-sdk-dev\bin\pub.bat run test:test --reporter=expanded
  'D:\apps\dart-sdk-dev\bin\pub.bat' is not recognized as an internal or external command,
  operable program or batch file.

failed with exit code 1
#0      runAsync (package:grinder/src/run.dart:108:9)
<asynchronous suspension>
#1      _RootZone.bindUnaryCallbackGuarded.<anon> (dart:async/zone.dart:1657:12)
<asynchronous suspension>

D:\code\dart\app_launcher>
```

I found the `lib\grinder_sdk.dart` need to update, change `pub ...` to `dart pub ...`.

devoncarew:
Fixed by https://github.com/google/grinder.dart/pull/386.

PR #386, opened 2022-03-31
try to fix pub.bat not found

liudonghua123:
see https://github.com/google/grinder.dart/issues/385.


devoncarew (approved):
Thanks for the contribution!

PR #388, opened 2022-06-07
bump to lints 2.0.0

pq:
/cc @devoncarew

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -16,7 +16,7 @@ dependencies:
   collection: ^1.15.0
 
 dev_dependencies:
-  lints: ^1.0.0
+  lints: ^2.0.0
   test: ^1.17.0
 
 executables:
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -16,7 +16,7 @@ dependencies:
   collection: ^1.15.0
 
 dev_dependencies:
-  lints: ^2.0.0
+  lints: '>=1.0.0 <3.0.0'
   test: ^1.17.0
 
 executables:
--- a/lib/src/grinder_task.dart
+++ b/lib/src/grinder_task.dart
@@ -50,20 +50,20 @@ class GrinderTask {
 
   /// This method is invoked when the task is started. If a task was created
   /// with a function, that function will be invoked by this method.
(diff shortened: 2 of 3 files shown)


pubspec.yaml
```diff
@@ -16,7 +16,7 @@ dependencies:
   collection: ^1.15.0
 
 dev_dependencies:
-  lints: ^1.0.0
+  lints: ^2.0.0
```
devoncarew, line 19: Looks like the CI is failing because it's running on Dart `2.16.0`, and lints 2 wants 2.17.

What I've done in the past - when I still want to have test coverage for an older sdk - is to broaden this dep to `>=1.0.0 <2.0.0`.

devoncarew: Separately, can lints `2.0.0` be installed on a 2.16 SDK? I'm not sure if it has lints which only work on 2.17 or not. I know we wanted to hold 2.0.0 until 2.17 was released, but I'm curious if now we can change the min sdk version to 2.16.

pq: Ah!  Great questions.  I'm not really sure here.

/fyi @mit-mit

mit-mit:
> Separately, can lints 2.0.0 be installed on a 2.16 SDK? I'm not sure if it has lints which only work on 2.17 or not. I know we wanted to hold 2.0.0 until 2.17 was released, but I'm curious if now we can change the min sdk version to 2.16.

My understanding what that some of the lints in 2.0 only just got added in 2.17. Isn't that correct @pq?

pubspec.yaml
mit-mit: > What I've done in the past - when I still want to have test coverage for an older sdk - is to broaden this dep to >=1.0.0 <2.0.0.

That seems like a fine solution. I believe that 2.0 is a strict superset of 1.0.

devoncarew: (sorry, above I meant `>=1.0.0 <3.0.0` - so provisioning either 1.x.x or 2.x.x would be ok)

pq: Sounds like a good approach.  I'll update.

devoncarew (approved):
👍 thanks!

PR #392, opened 2023-01-03
cmd line too long error on Windows

faithoflifedev:
Hi there,

On the Windows platform the grinder `analyze` task generates a "command line too long" error on Windows when there is a large number of files to analyze, since the code attempts to list all files that will be analyzed, rather then a folder name (it that's what has been provided  This PR resolves the issue on Windows, and I expect will not affect other platforms.


devoncarew (approved):
Thanks for the PR!

PR #428, opened 2026-06-22
Support cli_util 0.5.0

simolus3:
Declare support for `cli_util` versions 0.5.x. This package does not use `applicationConfigHome`, so it's not affected by breaking changes and can keep supporting 0.4.x too.


sethladd:
RSLGTM

